PR #2, opened 2015-12-08
WIP: Shared Memory data node

rohitvarkey:
`FileBlocks` compute node which can be computed to result in a `SharedMemory` data node.
The `SharedMemory` datanode creates a mmap array for the entire file, and splits the file to processes as per the offset ranges returned by the partition. Each process also has a mmap array open with the specified offsets (and length) and refs to these are stored with the pids. A gather on a `SharedMemory` returns the entire mmaped array.

`BytesPartition` returns offsets dividing the file on the basis of its filesize in terms of bytes. 

A `MapPartNode` on a `SharedMemory` results in a `DistMemory` with the results of the map operation stored at each chunk. A `gather` on a `DistMemory` with a `BytesPartition` results in a Array of the results at each process.

### DIFF
--- a/src/data-nodes/shared-memory/compute.jl
+++ b/src/data-nodes/shared-memory/compute.jl
@@ -9,7 +9,7 @@ immutable FileBlocks{P <: AbstractPartition} <: ComputeNode
     partition::P
 end
 
-FileBlocks(uri::AbstractString) = FileBlocks(uri, BytesPartition{nworkers()}())
+FileBlocks(uri::AbstractString) = FileBlocks(uri, BytesPartition())
 
 """
 Data Node for Shared Memory. (Dist File?)
@@ -38,9 +38,8 @@ Computing a FileBlock, will create a SharedMemory with the specified partition.
 """
 function compute(ctx, x::FileBlocks)
     targets = chunk_targets(ctx, x)
-    f = open(x.uri) #Opens a file handle to the file on the FileSystem
     #Reading them as bytes.
-    sharedarray = Mmap.mmap(f, Vector{UInt8}, filesize(f)-1, shared=true)
+    sharedarray = Mmap.mmap(x.uri, Vector{UInt8}, filesize(x.uri), shared=true)
     offsets = slice(ctx, x.uri, x.partition, targets) #Expects a vector of ranges
     refs = Pair[
         (targets[i] => remotecall(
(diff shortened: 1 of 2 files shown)


src/data-nodes/shared-memory/compute.jl
```diff
@@ -0,0 +1,97 @@
...
+function byte_splits(startpos, fsize, parts)
+    len = fsize - startpos
+    starts = len >= parts ?
+        round(Int, linspace(0, len, parts+1)) :
+        [[0:len;], zeros(Int, parts-len);]
+    map((x,y) -> x:y, starts[1:end-1], starts[2:end] .- 1)
+end
+
+function slice{n}(ctx, uri, ::BytesPartition{n}, targets)
+    f = open(uri)
```
tanmaykm: we shouldn't need to open the file to split. use `stat` instead?

rohitvarkey: Sure! Also, instead of doing `position`, it would be okay to take it as 0?

tanmaykm: Yes of course.

src/data-nodes/shared-memory/compute.jl
```diff
@@ -0,0 +1,97 @@
...
+
+function byte_splits(startpos, fsize, parts)
+    len = fsize - startpos
+    starts = len >= parts ?
+        round(Int, linspace(0, len, parts+1)) :
+        [[0:len;], zeros(Int, parts-len);]
+    map((x,y) -> x:y, starts[1:end-1], starts[2:end] .- 1)
+end
+
+function slice{n}(ctx, uri, ::BytesPartition{n}, targets)
```
tanmaykm: Also probably `slice` should not return an array. Not specific to this change, but in general.
The number of slices may be too large for a large entity. An iterator would be more efficient.

src/data-nodes/shared-memory/compute.jl
```diff
@@ -0,0 +1,97 @@
...
+    #Write to a file and return the file handle here?
+    n.sharedarray
+end
+
+##### Compute #####
+
+"""
+Computing a FileBlock, will create a SharedMemory with the specified partition.
+"""
+function compute(ctx, x::FileBlocks)
```
tanmaykm: Should `compute` always return `SharedMemory` datanodes?
How does this work when the workers are not on same machine?

shashi: Would mmap-ing a file from HDFS do the trick?

tanmaykm: Nope. HDFS blocks will be spread across machines.
I think making `FileBlocks` a `DataNode` would be the right approach.

PR #4, opened 2015-12-19
Implementing BFS over ComputeFramework

pranavtbhat:
This is an attempt at implementing BFS. I managed to get map to work, but I got a convert error on reduce and filter went into a stackoverflow. Not sure what I'm doing wrong.

Design issues I've faced are:

- The datastructure to store the graph. @ViralBShah recommended a sparse matrix, but I've found a list of lists to work better for BFS. The graph needs to be accessible by all worker processes. One strategy is to use `DistributedArrays`, and the second is to use `SharedArray`. However I couldn't create a SharedArray of type `Vector{Int}` (it requires a bits type) .
- The distance vector must be available to all worker processes, and be write synchronized. I implemented this as a SharedArray. But how do I reference this array in the worker processes (inside `filterBFS`).
- Do I use a single seed vertex or can I use multiple? A single seed vertex wastes the first iteration.

### DIFF
--- a/examples/bfs.jl
+++ b/examples/bfs.jl
@@ -1,0 +1,60 @@
+
+addprocs(2)
+
+# Map function for BFS. Returns all vertices reachable from the input vertex
+@everywhere function mapBFS(t)
+  u = t[1]
+  d = t[2]
+  [(v,d+1) for v::Int in find(sparseMatrix[u,:])]
+end
+
+# Filter function for BFS. Discard already-explored vertices
+@everywhere function filterBFS(tuple)
+  u,d = tuple
+  if dist[u] < 0
+    dist[u] = d
+    return true
+  else
+    return false
+  end
+end
+
+
+using ComputeFramework
+
+@everywhere numVertices = 8
+@everywhere adjacencyMatrix =  [ 0  1  1  1  0  0  0  0;
+                                 1  0  1  0  1  1  0  0;
+                                 1  1  0  1  0  1  1  0;
+                                 1  0  1  0  0  0  1  0;
+                                 0  1  0  0  0  1  0  1;
+                                 0  1  1  0  1  0  1  1;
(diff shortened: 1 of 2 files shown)


shashi:
> The datastructure to store the graph. @ViralBShah recommended a sparse matrix, but I've found a list of lists to work better for BFS.

There's actually little difference between the two representations - a column in a sparse matrix is a list of non-zero values in the column - which boils down to an adjacency list.

This is a good start, however, I think the point of this exercise is to introduce SparseMatrix layout type(s) which can split and distribute a sparse matrix, just like `distribute(array, RowLayout())` splits an array by rows and puts parts on all the processors. Next steps would be to see if we can use the framework to work through BFS, if not, what is missing, and where should they be implemented according to the design. As a good rule of thumb, sorting, distribution or redistribution of data should be implemented as Layouts (used with `distribute` and `redistribute` nodes), high-level APIs for algorithms should be implemented as ComputeNodes.

pranavtbhat:
This BFS example uses Map Reduce. This is not efficient and relies on synchronized and shared memory. I've decided to implement it anyways so that we can compare performance later.

Bulk Synchronous Parallel model is the way to go. This is similar to what Apache Giraph(Pregel) and GraphX use. Luckily GraphX is in its alpha stage and Giraph is kind of slow (its built on directly on Hadoop while GraphX is build on Spark). 

To implement BSP, we need a form of message passing. I found [this](http://blog.acolyer.org/2015/05/26/pregel-a-system-for-large-scale-graph-processing/) article extremely informative. Implementing BFS is pretty straightforward if we use this model.

ViralBShah:
Also see the GunRock papers for BFS on GPUs.

pranavtbhat:
The Map-Reduce BFS is working. I've added tests for `filter` and `reduce` as well.

pranavtbhat:
I'll move this to a different branch.

PR #11, opened 2015-12-24
WIP: hdfs file node

tanmaykm:
*do not merge*

First attempt at getting HDFS files to work with ComputeFramework.

Couldn't avoid quite a bit of duplication, I think data nodes need further generalization. Also, we should have extensions like this be built as separate packages. Will try these next.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,3 @@
 julia 0.4
 Compat
+Elly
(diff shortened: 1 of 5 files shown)


shashi:
This is looking good!

I think we can make SplitNode take either a FileNode or a HDFSFileNode and still work with the same compute method.

I'll try to do this now.

shashi:
Can you take a look?

shashi:
This is very obsolete by now. Let's start over.

ViralBShah:
Yes - worth getting it in.

PR #14, opened 2015-12-29
TypeLayouts

pranavtbhat:
I've created this basic graph representation that should be compatible with other Julia graph libraries such as LightGraphs.jl and Graphs.jl.  The representation is very light; other auxillary structures such as labels etc. should be passed into functions separately. 

The graph supports both Adjacency List and Adjacency Matrix (SparseMatrixCSC also). I don't anticipate using `redistribute` on the GraphLayout though.


src/graph-layouts.jl
```diff
@@ -0,0 +1,89 @@
...
+###
+
+# Allowed Graph datastructures:
+typealias AdjacencyList Array{Array{Int, 1}, 1}
+typealias AdjacencyMatrix AbstractArray{Int, 2}                      # Covers sparse matrices
+typealias GraphStruct Union{AdjacencyList, AdjacencyMatrix}
+
+
+# Generic (undirected or directed, up to the user to decide) Graph
+type DistGraph{S<:GraphStruct}
```
shashi: why is this needed? `distribute` combined with  `GraphLayout` can do the job right?

src/graph-layouts.jl
```diff
@@ -0,0 +1,89 @@
...
+###
+# Graph layouts
+###
+
+immutable GraphLayout{S<:GraphStruct} <: AbstractLayout end
+
+"""
+Given an input vertex range, slice it into sub-ranges.
+"""
+function slice(ctx, vrange::UnitRange{Int}, ::GraphLayout, targets)
```
shashi: This function should be named something else.

src/graph-layouts.jl
```diff
@@ -0,0 +1,89 @@
...
+# Allowed Graph datastructures:
+typealias AdjacencyList Array{Array{Int, 1}, 1}
+typealias AdjacencyMatrix AbstractArray{Int, 2}                      # Covers sparse matrices
+typealias GraphStruct Union{AdjacencyList, AdjacencyMatrix}
+
+
+# Generic (undirected or directed, up to the user to decide) Graph
+type DistGraph{S<:GraphStruct}
+    nv::Int                            # not really required, but kept for convenience
+    vertices::UnitRange{Int}
```
shashi: So far I don't see why we need this field.

Am thinking of allowing `slice` to return some metadata about each chunk and then storing that metadata in the DistMemory. So that each layout type can put some metadata along with the chunks to facilitate things like `length`, `getindex

pranavtbhat: The vertices field denotes the range of vertices present locally. It allows the workers to determine the global index of the first worker they have. But I guess the metadata will do something similar.

src/graph-layouts.jl
pranavtbhat: I'll change the name.

src/graph-layouts.jl
pranavtbhat: This representation is similar to what Graphs.jl and LightGraphs.jl use. Implementing `convert` methods will allow users to apply algorithms directly on their graphs. Writing algorithms will also be more convenient with a representation such as this. 

Will this cause any performance problems?

pranavtbhat:
I had an idea. Instead of having a graph layout, how about we have a generic `TypeLayout`?
Suppose I need to distribute this an object of this type:
```julia
type complex
   f1::Vector
   f2::AbstractArray{Any,2}
end
```
We can have then have a TypeLayout that describes the layouts of `f1` and `f2`:

```julia
type TypeLayout{T<:Layout}
   layouts::Vector{T}
end

obj = complex(ones(10), ones(10,10))
distribute(obj, TypeLayout([RowLayout(), ColumnLayout()]))
```
Implementing `slice` and `gather` should then be easy and similar to what I've written in `graph-layouts.jl`.

pranavtbhat:
This commit contains the generic TypeLayout definition as well as generic slice and gather methods. 

Can we also have a SingletonLayout? Sometimes I need to pass a single value into a map (a common argument). For example:
```julia
multiply_by(x,y) = x*y
result = gather(Context(), map(multiply_by, distribute(rand(10)), distribute(5, SingletonLayout{Int}())))
```
This layout will simply produce an array of the same element, and slice it.

src/ComputeFramework.jl
```diff
@@ -76,6 +76,8 @@ include("file-nodes.jl")
 include("redistribute.jl")
 include("sort.jl")
 
+include("graph-layouts.jl")
```
shashi: File doesn't exist anymore.

shashi:
> Can we also have a SingletonLayout?

how is it different from `distribute` with `Bcast` layout?

src/ComputeFramework.jl
pranavtbhat: Oops sorry.

pranavtbhat:
Oh I hadn't checked BroadcastLayout. Never mind its the same thing.

pranavtbhat:
I guess metadata solves this problem.

PR #20, opened 2016-02-19
Use isa instead of typeof in show

andreasnoack:
Couldn't show the `mapredure` from the readme.

I couldn't run the test on either 0.4 and 0.5, but I had to give up on fixing that. It might be a good idea to enable Travis testing and code coverage.

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -85,7 +85,7 @@ end
 
 
 function show(io::IO, x::MapReduce)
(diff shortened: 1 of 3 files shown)


shashi:
Thanks, I am working on the `rewrite` branch and factoring out Chunks into a separate repo. Will put in Travis once I merge it.

PR #23, opened 2016-03-31
Update REQUIRE to include Logging

ivirshup:
Trying to use the package if Logging isn't installed raises a LoadError.


shashi:
Thanks :)

Issue #21, opened 2016-03-29
stage functions should not rely on the domain being a DomainBranch


pranavtbhat:
```julia
a = compute(Distribute(BlockPartition(10, 1), spzeros(10, 10)))
da = compute(a*a)
compute(map(x->1, da))
```

This throws an error:
```
ERROR: type DenseDomain has no field children
 in stage at /home/pranav/.julia/v0.5/ComputeFramework/src/map-reduce.jl:13
 [inlined code] from /home/pranav/.julia/v0.5/ComputeFramework/src/compute.jl:50
 in compute at /home/pranav/.julia/v0.5/ComputeFramework/src/compute.jl:51
 in eval at ./boot.jl:265
```
Other operations on `da` still work.

pranavtbhat:
Fixed.

shashi:
Huh? how?

This is still an issue, `stage` methods should not rely on the domain being a `DomainBranch`.

shashi:
it's an easy fix. We need to have a `children` generic function as opposed to `d.children` which returns a 1-element domain in case of a non-DomainBranch object.

PR #24, opened 2016-04-06
Added generic children function

pranavtbhat:
I've added a generic children function that returns the `children` field for `DomainBranch` and indices for the other `Domain` sub-types.

This should prevent `stage` from relying on all Domain sub-types having a `children` field. (#21)

### DIFF
--- a/src/domain.jl
+++ b/src/domain.jl
@@ -103,6 +103,8 @@ isempty(a::DomainBranch) = isempty(head(a))
 intersect{D<:Domain}(a::DomainBranch{D}, b::D) = intersect(head(a), b)
 project{D<:Domain}(a::DomainBranch{D}, b::D) = intersect(head(a), b)
 getindex{D<:Domain}(a::DomainBranch{D}, b::D) = getindex(head(a), b)
+children(x::DomainBranch) = x.children
+
 
 
 ###### Array Domains ######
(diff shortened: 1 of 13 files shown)


src/domain.jl
```diff
@@ -120,6 +122,7 @@ DenseDomain(xs...) = DenseDomain(xs)
 DenseDomain(xs::Array) = DenseDomain((xs...,))
 
 indexes(a::DenseDomain) = a.indexes
+children(a::DenseDomain) = a.indexes
```
shashi: This should just return a

src/sparse.jl
```diff
@@ -9,6 +9,7 @@ end
 size(d::SparseCSCDomain, arg...) = size(d.adomain, arg...)
 length(d::SparseCSCDomain) = Int(d.colptr[end]-1)
 indexranges(d::SparseCSCDomain) = indexranges(d.adomain)
+children(d::SparseCSCDomain) = d.adomain
```
shashi: I thought I deleted this file.... haha anyway this thing needs a second look. We will need sparsecscdomain back like i mentioned in the last email

shashi:
I forgot to mention this before. We need a `parts` generic function defined on `AbstractPart` (defaults to identity fn), and `Cat` for which it is`x.parts`, and use `parts(x)` wherever there is `x.parts`.

If you can get travis to pass I'll merge this.

pranavtbhat:
I've added generic accessors for `children` and `parts`. Rebased as well. Should I replace occurrences in the tests as well?

shashi:
yes

pranavtbhat:
Done.

PR #25, opened 2016-04-11
Tweaks

MikeInnes:
Just a couple of small tweaks I made reading through the package. I'm still getting to grips so let me know if anything was intentional.


shashi:
Awesome! Thanks!

Issue #13, opened 2015-12-26
compute on MapReduce and Reduce returns gathered value


pranavtbhat:
Running `compute` on `MapReduce`/`Reduce` nodes gives the final gathered value. Wouldn't it be preferable to return a `DataNode` instead?

shashi:
Reduce usually results in a value, so it made sense to just give the final result. But now I'm not so sure.

It's possible to move the final reduce step (on the parent) to the `gather` on a reduce node. I agree that would be more consistent and less confusing.

pranavtbhat:
Since Map and Reduce are implemented separately, I guess this issue is resolved. Maybe we can move the keyword argument `get_result` from `reduceblock` to [reduce](https://github.com/shashi/ComputeFramework.jl/blob/master/src/map-reduce.jl#L49) instead?

shashi:
`reduce` functions now just return the answer.

PR #28, opened 2016-04-24
WIP: Map-reduce

pranavtbhat:
Fixed the `mean` method. Also added the `mapreduce` method. 

The `count` example seemed obsolete since it used an Accumulator, so I replaced it with a map-reduce example. 

Should I move the keyword argument as mentioned in #13 ? I can also write tests for `map-reduce.jl` if required.


shashi:
Next time, do split the commits.

I don't fully understand what would happen if I set `get_result=false` in `reduce`?

pranavtbhat:
I think only the metadata will be returned when the thunk is computed. But since most applications of `reduce` will just result in a value, I think `get_result` can just be set to `true`.

Moving the keyword argument to the exported `reduce` function will give the user most flexibility I guess.

shashi:
We should just export reduceblock

PR #32, opened 2016-04-25
GPU support for ComputeFramework

ranjanan:
Add GPU support using ArrayFire

### DIFF
--- a/src/array/alloc.jl
+++ b/src/array/alloc.jl
@@ -10,7 +10,7 @@ function stage(ctx, a::AllocateArray)
     branch = partition(a.partition, a.domain)
     dims = length(indexes(a.domain))
     alloc = let eltype = a.eltype, f = a.f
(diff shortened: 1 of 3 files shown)


shashi:
What do you think we should do about the ArrayFire dependency? probably [`@require`](https://github.com/MikeInnes/Requires.jl)?

can you create a branch on this repo? I can then add my commits

ranjanan:
Yeah, `@require` seems like it's worth trying out. Shall I delete this pull request and create a branch on the repo, which we can then collab on?

shashi:
yea sounds good

ViralBShah:
Doesn't `@require` will break precompilation.

shashi:
On 0.4 at least if you don't have precompile directive, it does work.

The alternative is having users compile ArrayFire to get CF to work, or have a package for interop. Another case for https://github.com/JuliaLang/julia/issues/15705

PR #33, opened 2016-04-26
WIP: GPU Support for ComputeFramework

ranjanan:
This PR aims to allow ComputeFramework to use GPUs by integrating with [`ArrayFire.jl`](https://github.com/JuliaComputing/ArrayFire.jl).

### DIFF
--- a/src/compute.jl
+++ b/src/compute.jl
@@ -311,11 +311,6 @@ end
 _move(ctx, to_proc, x) = x
 _move(ctx, to_proc::OSProc, x::AbstractPart) = gather(ctx, x)
 
-@require ArrayFire begin 
-    using ArrayFire
-    _move(ctx, to_proc::GPUProc, x::AbstractPart) = AFArray(gather(ctx, x))
-end
-
 function do_task(ctx, proc, thunk_id, f, data, chan, send_result)
     try
         res = f(map(x->_move(ctx, proc, x), data)...)
@@ -330,8 +325,3 @@ function async_apply(ctx, p::OSProc, thunk_id, f, data, chan, send_res)
     remotecall(do_task, p.pid, ctx, p, thunk_id, f, data, chan, send_res)
 end
 
-@require ArrayFire begin
-    function async_apply(ctx, p::GPUProc, thunk_id, f, data, chan, send_res)
-        remotecall(do_task, p.pid, ctx, p, thunk_id, f, data, chan, send_res)
-    end
-end
--- a/src/compute.jl
+++ b/src/compute.jl
@@ -311,15 +311,27 @@ end
 _move(ctx, x) = x
 _move(ctx, x::AbstractPart) = gather(ctx, x)
 
+@require ArrayFire begin 
+    using ArrayFire
+    _move(ctx, to_proc::GPUProc, x::AbstractPart) = AFArray(gather(ctx, x))
+end
+
 function do_task(ctx, proc, thunk_id, f, data, chan, send_result)
     try
-        res = f(map(x->_move(ctx, x), data)...)
+        res = f(map(x->_move(ctx, proc, x), data)...)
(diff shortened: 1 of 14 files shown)


src/compute.jl
```diff
@@ -311,15 +311,27 @@ end
...
 _move(ctx, x::AbstractPart) = gather(ctx, x)
 
+@require ArrayFire begin 
+    using ArrayFire
+    _move(ctx, to_proc::GPUProc, x::AbstractPart) = AFArray(gather(ctx, x))
+end
+
 function do_task(ctx, proc, thunk_id, f, data, chan, send_result)
     try
-        res = f(map(x->_move(ctx, x), data)...)
```
shashi: ? o_O

I'll fix

shashi: Oh, actually nothing's wrong.

shashi:
I can't make sense of the travis failure on 0.5, it looks like it is due to MacroTools, however there is no use of MacroTools in this package.

shashi:
@tkelman do you know what's up?

tkelman:
You should **not** use Requires.jl. It's a nasty hack with zero tests that hooks into things it shouldn't be touching, and interacts badly with precompilation.
```
ERROR: LoadError: LoadError: InitError: type Method has no field func
 in eval(::Module, ::Any) at ./boot.jl:226
 in eval(::Module, ::Expr) at /home/travis/.julia/v0.5/Requires/src/Requires.jl:3
 [inlined code] from /home/travis/.julia/v0.5/Requires/src/require.jl:9
 in (::Requires.##8#9)() at /home/travis/.julia/v0.5/Requires/src/init.jl:17
 in __init__() at /home/travis/.julia/v0.5/Requires/src/init.jl:24
 in _require_from_serialized(::Int64, ::Symbol, ::String, ::Bool) at ./loading.jl:174
 in require(::Symbol) at ./loading.jl:365
 in include_from_node1(::String) at ./loading.jl:426
 in eval(::Module, ::Any) at ./boot.jl:226
 [inlined code] from ./sysimg.jl:11
 in require(::Symbol) at ./loading.jl:351
 in include_from_node1(::String) at ./loading.jl:426
 in process_options(::Base.JLOptions) at ./client.jl:263
 in _start() at ./client.jl:319
during initialization of module Requires
while loading /home/travis/.julia/v0.5/ComputeFramework/src/ComputeFramework.jl, in expression starting on line 5
```

shashi:
Ah, that's the problem. Thanks! Well the alternative is to move GPU support into a new 20-line package. Can't possibly ask users to set up ArrayFire to use this package.

tkelman:
Or make ArrayFire depend on ComputeFramework. Isn't ComputeFramework pure julia and relatively small?

shashi:
That's an option. @ranjanan? (I believe both packages are not registered yet)

ranjanan:
ArrayFire isn't registered now. I'm spending some time making ArrayFire work with the C interface, so it'll work with v0.4. Then we can register it and put it in one of our REQUIRE files.

jpsamaroo:
So it's been a few years since this PR was posted, and surely it's a bit out-of-sync with master. @ranjanan and @shashi , do either of you still have desire to see this sort of functionality added to Dagger?

I personally feel that GPU support is somewhat outside of the realm of Dagger, especially since we now have excellent CUDA support and WIP ROCm support in Julia (I assume this PR was created back when CuArrays didn't exist/wasn't ready for use). I also don't know how the scheduler would be able to efficiently allocate work to GPU nodes without having information about the number and capabilities of GPUs on each worker, and I don't think it wise to add such support to Dagger.

Maybe an alternative design would be to let the user pass information to the scheduler about which workers might have special hardware like GPUs, and which thunks are suitable to execute on such hardware; the scheduler can then preferentially move such thunks to matching workers. I'd of course be happy to drive the work for such a design if it's desired.

jpsamaroo:
If I don't hear any objections, I'm going to close this PR next week.

PR #34, opened 2016-04-30
WIP: Message Passing

pranavtbhat:
Message Passing between processes. The example shows the basic producer-consumer example. The message passing will be useful for cooperative tasks. 

It will also enable the bulk synchronous parallel model, as this will allow tasks to be executed in parallel over worker processes, which can co-operate through message passing, in a synchronized manner.

### DIFF
--- a/examples/mpass.jl
+++ b/examples/mpass.jl
@@ -1,0 +1,22 @@
+addprocs(2)
+
+using ComputeFramework
+
+@everywhere function transmit(id)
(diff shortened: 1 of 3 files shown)


ViralBShah:
I am not sure I like introducing the lower level abstraction of processes and message in here. We should probably do a call next week.

pranavtbhat:
Sure. I'll try to get some pseudocode ready by then.

PR #35, opened 2016-05-12
Design Discussion: API

MikeInnes:
This is the prototype for an experimental new programming model for ComputeFramework.

Although we discussed this in the context of Flow.jl, I've actually stripped it out from here since I wanted to distill the various issues as much as possible. So this is the simplest hack that demonstrates what I'd like the frontend to look like.

### Background

Right now you use CF by building expression trees and evaluating them. It's analogous to working with Julia code like this:

```julia
compute(x) = eval(x) # or some custom interpreter
f(x) = :($x^2)
g(xs) = :(sum([$(map(x -> f(x), xs)...)])

g(1:10) == :(sum([1 ^ 2,2 ^ 2,3 ^ 2 ... ]))
compute(g(1:10)) == 385
```

Some think of this as "lazy" or "symbolic" programming, but I hope this illustrates why I disagree; Julia's semantics didn't change just because we called `compute` explicitly. Likewise, CF's DSL is both strict and non-symbolic, it just doesn't (yet) have proper syntax. `</nitpicking>`

What I'd like to do is move from "build expression and eval" to "write a function, take some data, run `f(x)`. This has a number of benefits (including making a few things easier on the developer side), but first and foremost is that it's potentially much more intuitive, since the semantics and programming model are very close to the host language. Let me explain:

### Implementation

[at the moment this is mainly illustrative, the implementation has some issues due to `getindex` not working in CF.]

This PR introduces the `CFArray` type. `CFArray` is just a regular ol' piece of data, and as an `AbstractArray` you can use the same functions you know and love. Of course, the data and operations on it are all parallelised.

```julia
xs = rand(CFArray, 10^4, 10^4)
10000×10000 CFArray{Float64,2}:
 0.315912  …  0.42977
 ⋮         ⋱         
 0.484693     0.69204

julia> exp(xs)
10000×10000 CFArray{Float64,2}:
 1.51383  …  1.42473
 ⋮        ⋱         
 1.03595     1.59556

julia> xs'*xs
10000×10000 CFArray{Float64,2}:
 3321.84  …  2514.66
    ⋮     ⋱         
 2514.66     3328.39

mvgavg(xs,n) = (xs[1:end-n]+xs[n+1:end])/n

julia> mvgavg(rand(CFArray, 100), 2)
98-element Array{Float64,1}:
 0.504985
 ⋮       
 0.55807
```

Notice how many fewer concepts are required to get going.

The main issue here is that the operations within `mvgavg` aren't visible to CF's runtime, which means they'll execute in sequence (though each one will be parallelised in itself – like sequential calls to BLAS). CF may be able to apply more optimisations if it has a view of the whole program graph, and to enable this we use the `@cf` macro:

```julia
@cf mvgavg(xs,n) = (xs[1:end-n]+xs[n+1:end])/n
```

When `mvgavg` is called with a `CFArray`, the function will build up a larger DAG and ask CF to run it – all in a way which is transparent to the user. `@cf` functions can also be used to extend existing functions, and can work with standard Julia data types like `Array` as well.

Pretty straightforward, I think – but of course, that's the point. What do people think?

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -10,6 +10,7 @@ export CFArray, @cf
 type CFArray{T,N} <: AbstractArray{T,N}
   head::Computation
 end
+@forward CFArray.head size
 
 function CFArray(xs::Computation)
   xs = compute(xs)
@@ -20,12 +21,12 @@ macro cf(ex)
   @capture(shortdef(ex), name_(arg_) = exs__) ||
     error("invalid @cf function")
   quote
-    ex
-    $name($arg::ComputeFramework) = ComputeFramework.CFArray($name($arg.head))
+    $ex
+    $name($arg::ComputeFramework.CFArray) = ComputeFramework.CFArray($name($arg.head))
   end |> esc
 end
 
-Base.getindex(xs::CFArray, args::Integer...) = compute($f($xs.head, args...))
+Base.getindex(xs::CFArray, args::Integer...) = compute(getindex(xs.head, args...))
 
 for f in :[getindex map reduce exp log].args
   @eval $f(xs::CFArray, args...) = CFArray($f(xs.head, args...))
--- a/src/api.jl
+++ b/src/api.jl
@@ -5,7 +5,7 @@ import Base: eltype, ndims
 export CFArray, @cf
 
 @forward Computed.result eltype, ndims
-@forward Cat.parttype eltype, ndims
+ndims(x::AbstractPart) = ndims(domain(x))
 
 type CFArray{T,N} <: AbstractArray{T,N}
   head::Computation
@@ -26,7 +26,9 @@ macro cf(ex)
   end |> esc
 end
 
-Base.getindex(xs::CFArray, args::Integer...) = compute(getindex(xs.head, args...))
+# x = CFArray(rand(BlockPartition(100), 10^3))
+
+Base.getindex(xs::CFArray, args::Integer...) = @> getindex(xs.head, args...) gather first
 
 for f in :[getindex map reduce exp log].args
   @eval $f(xs::CFArray, args...) = CFArray($f(xs.head, args...))
@@ -36,7 +38,7 @@ for f in :[(+) (*)].args
   @eval $f(xs::CFArray, ys::CFArray) = $f(xs, ys)
 end
 
-for f in :[(.+) (.*) ].args
+for f in :[(.+) (.*)].args
   @eval $f(xs::CFArray, ys::Number) = $f(xs, ys)
   @eval $f(xs::Number, ys::CFArray) = $f(xs, ys)
   @eval $f(xs::CFArray, ys::CFArray) = $f(xs, ys)
--- a/src/api.jl
+++ b/src/api.jl
@@ -28,7 +28,9 @@ end
 
 # x = CFArray(rand(BlockPartition(100), 10^3))
 
-Base.getindex(xs::CFArray, args::Integer...) = @> getindex(xs.head, args...) gather first
+Base.collect(xs::CFArray) = gather(xs.head)
+
+Base.getindex(xs::CFArray, args::Integer...) = @> xs.head getindex(args...) gather first
 
 for f in :[getindex map reduce exp log].args
   @eval $f(xs::CFArray, args...) = CFArray($f(xs.head, args...))
--- a/src/api.jl
+++ b/src/api.jl
@@ -1,0 +1,42 @@
+using MacroTools, Lazy
+
+import Base: eltype, ndims
+
+export CFArray, @cf
+
+@forward Computed.result eltype, ndims
+@forward Cat.parttype eltype, ndims
+
+type CFArray{T,N} <: AbstractArray{T,N}
+  head::Computation
+end
+
+function CFArray(xs::Computation)
+  xs = compute(xs)
+  CFArray{eltype(xs),ndims(xs)}(xs)
+end
+
+macro cf(ex)
+  @capture(shortdef(ex), name_(arg_) = exs__) ||
+    error("invalid @cf function")
+  quote
+    ex
+    $name($arg::ComputeFramework) = ComputeFramework.CFArray($name($arg.head))
+  end |> esc
+end
+
+Base.getindex(xs::CFArray, args::Integer...) = compute($f($xs.head, args...))
+
+for f in :[getindex map reduce exp log].args
+  @eval $f(xs::CFArray, args...) = CFArray($f(xs.head, args...))
+end
+
+for f in :[(+) (*)].args
+  @eval $f(xs::CFArray, ys::CFArray) = $f(xs, ys)
+end
+
+for f in :[(.+) (.*) ].args
+  @eval $f(xs::CFArray, ys::Number) = $f(xs, ys)
+  @eval $f(xs::Number, ys::CFArray) = $f(xs, ys)
+  @eval $f(xs::CFArray, ys::CFArray) = $f(xs, ys)
+end
--- a/src/ComputeFramework.jl
+++ b/src/ComputeFramework.jl
@@ -28,6 +28,7 @@ include("array/operators.jl")
 include("array/getindex.jl")
 include("array/matrix.jl")
 
+include("api.jl")
 include("array/show.jl")
 
 end # module
--- a/src/array/getindex.jl
+++ b/src/array/getindex.jl
@@ -5,6 +5,9 @@ end
 
 Base.getindex(c::Computation, idx...) = GetIndex(c, idx)
 
+Base.size(c::Computed, i::Int) = size(domain(c.result),i)
+Base.size(c::Computed) = size(domain(c.result))
+
 function stage(ctx, gidx::GetIndex)
     inp = cached_stage(ctx, gidx.input)
     dmn = domain(inp)


datnamer:
Does the Cf macro parallelise all the way down the call stack of any  arbitrary function?

MikeInnes:
In a word, yes.

For example, calling the `mvgavg` function transparently constructs and evaluates a DAG under the hood, and that DAG includes all of the `getindex` internals and so on.

When you say "arbitrary function", you may be referring to unadorned Julia functions like `f(x) = x+x`. In a real implementation, only `@cf` adorned functions will be spliced in, and others would run independently.

Is this a big loss? I personally think that the superficial benefit of writing unadorned functions like the above is just that, superficial – the set of functions like `f` that will work with both `Array` and `Computation` is very small, and notably, trying to use any kind of control flow, loops, or `@show`-debugging will break the expression-building API in unintuitive ways. (Not that these are common, right?) This is the primary reason I don't like it – it's fundamentally very brittle.

The programming model I've outline above ensures that a function of `AbstractArray` will *always* work correctly with `CFArray`, even if it isn't always as fast – and even that is easily fixed. Aside from being a much less leaky abstraction, the macro approach in principle allows for more advanced things like digging into the source of a dependent function written for Julia functions and splicing them into the DAG, which would remove this downside entirely.

datnamer:
Interesting.

So "control flow, loops, or @show-debugging" all work with cf functions?  Super cool.

Does this automagically transfer to out of core and distributed backends?

src/api.jl
```diff
@@ -0,0 +1,42 @@
...
+  xs = compute(xs)
+  CFArray{eltype(xs),ndims(xs)}(xs)
+end
+
+macro cf(ex)
+  @capture(shortdef(ex), name_(arg_) = exs__) ||
+    error("invalid @cf function")
+  quote
+    ex
+    $name($arg::ComputeFramework) = ComputeFramework.CFArray($name($arg.head))
```
shashi: what should `::ComputeFramework` be instead?

MikeInnes: Oh whoops. `CFArray`.

shashi: Ah, it's CFArray

shashi:
Thanks for the prototype! This is very interesting.

The change boils down to:

1. `CFArray`s are `compute`d on creation
2. If a function is annotated with `@cf`, then the result of the function is only `compute`d at the end of the function; if it's not annotated with `@cf`, it would `compute` each operation step-by-step.

Firstly, this is better for exploratory coding, since you don't need to call `compute`, and can play with other AbstractArrays on your functions. Secondly, this means `CFArray` will work with any function defined on AbstractArrays. However you will need to annotate `@cf` at the beginning of the function to take advantage of CF's optimizations.

cc @ViralBShah who was unsure about the implications of this.

I have a few questions:

1. Would this sit on top of `Computation` as it does now? Wrapper types are a pain because you need to define all the methods on them all over again. I'd love to have the gross change remove a couple of hundred lines of code instead of adding to it. ;) I'll think more about how a DAG output from this macro can be walked to create the granular DAG input to scheduler.
2. What is then the way to pass in a `Context`? can I have a `@cf` function where I use many contexts?
3. Can you elaborate on how control-flow would work? Call `compute` at the end of each iteration? Sometimes it needs to happen in the [middle of the loop body](https://github.com/shashi/ComputeFramework.jl/blob/master/examples/logreg.jl#L24). Here, for example `cholfact!` will be really slow if you don't notice that you need to get the thing to a single process.

I feel a hint of sadness about stopping to expose the actual model i.e. having the user construct a `Computation`, and call `compute` with a `Context`. It is less fancy and more true to what is happening and kind of popular too because of things like TensorFlow. (We should keep it if it's easy to.)

However, I see the appeal in this:

> the set of functions that will work with both Array and Computation is very small, and notably, trying to use any kind of control flow, loops, or @show-debugging will break the expression-building API in unintuitive ways. (Not that these are common, right?) This is the primary reason I don't like it – it's fundamentally very brittle.

I don't see it as brittle though, just low-level and outsider-like.

Lastly, I don't think `@cf` is the best name for this thing. `@fuse` or `@collapse` or `@optimize` or `@compact` or even `@flow` are candidates. given

> @cf functions can also be used to extend existing functions, and can work with standard Julia data types like Array as well.

MikeInnes:
To clarify a little about the control-flow stuff, this is more of an "in-principle" thing right now. Of course having `if` and `for` syntax is a little redundant without a backend to run it :) Figuring out how those things should work within CF is a big discussion on its own, but I don't think it's controversial that they'd be useful in some form.

Also, it occurs to me that there is actually a precedent for this kind of API – just look a [Numba](http://numba.pydata.org) for Python. You can use Numba in your sleep, just take a regular python function and add `@jit` so Numba can run custom optimisations on it. I think `@cf` could be pretty much the same, albeit with very different goals (and I only really use that name as a placeholder, I'm sure we can come up with something better).

@shashi, to your questions:

1. Yep, I definitely think choosing one API and sticking to it is the way forward. We could replace a big chunk of CF's code with generic operations from Flow.
2. For one thing you could auto-generate `f(::Context, args...)` methods. When Julia methods call ComputeFramework ones you don't want the Context argument to leak all the way up the call stack, so dynamic binding might be a good idea. We don't use that in Julia much yet but it's often used to great effect in Clojure for this kind of problem. (that issue isn't specific to this api, though)
3. Briefly, I don't think CF should handle that kind of for loop at all – Julia's handling it just fine. What'd be more interesting is writing a `for` loop over distributed `CFArray`s and having it appropriately parallelised and optimised – like the tensor contraction stuff we talked about.

> I feel a hint of sadness about stopping to expose the actual model ... it is less fancy and more true to what is happening

I definitely understand your concern here, and transparency is often an important goal. At the same time, I think there's a strong precedent for the user's mental model being different from what's "really" happening. Any other time you program – whether via numba, Julia, whatever – has the same set of steps, build an expression tree, apply optimisations and lowering, invoke a runtime, but these are all transparent to the user. Does this mean they aren't true to what's happening, or does it mean they're abstracting over implementation details?

An abstraction can certainly be a bad thing if you end up having to break through it to get stuff done. For example, if you weren't easily able to express things like custom `Contexts`, a point you've raised above. I think we can test this easily enough by sketching out how the API looks with gradually more complex problems.

shashi:
> We could replace a big chunk of CF's code with generic operations from Flow.

What would an example generic operation look like? As a concrete example, how would it represent say matvec multiply or transpose.

> so dynamic binding might be a good idea.

+1

> I don't think CF should handle that kind of for loop at all

Ah, I didn't see that. So how to express that for loop is an unsolved question for now...

> You can use Numba in your sleep, just take a regular python function and add @jit so Numba can run custom optimisations on it.

What are regular python functions? Do they have for loops? Or use higher level functions like `map` & `reduce`?

MikeInnes:
> What would an example generic operation look like?

To clarify (terminology is hard :) I'm thinking about compiler-level operations (like modifying the graph) rather than user level operations (which would be nodes within the graph). I just meant stuff like constructing the graph, resolving variables, CSE, various kinds of inference and analysis and so on.

> So how to express that for loop is an unsolved question for now...

I could definitely see a future in which CF is smart enough to do gathers and run single-core versions if it would be faster... though that's a little way off. For now it doesn't really matter since you can choose where CF and Julia each run.

> What are regular python functions? Do they have for loops? Or use higher level functions like map & reduce?

I think in principle any Python is valid, although from the looks of it Numba is targeted at optimising C-style loops and arithmetic – not sure what else it can handle.

shashi:
>> So how to express that for loop is an unsolved question for now...
>
>I could definitely see a future in which CF is smart enough to do gathers and run single-core versions if it would be faster... though that's a little way off. For now it doesn't really matter since you can choose where CF and Julia each run.

Is it possible to allow the use of gather inside a `@cf ` function? I think we can get this API right now if we can allow this... I am convinced now that we need this API, but would like a middle ground for people manually wanting to do gather etc (or for when you need it: a way to go from CF array (or Julia array - to keep code still generic) to Julia array). `compute` does clutter real-world code, especially notebooks...

MikeInnes:
So me and Shashi have been discussing this and came up with an alternative way to get the array interface without as much of the macrology.

So basically the idea is to make `CFArray` a lazy data structure which waits as late as possible to realise itself, but still leaves this process implicit. If you write:

```julia
Z = X'Y
```

where `X` and `Y` are `CFArray`s, then `Z` is going to return immediately with a new `CFArray` which represents the computation (using ComputeFramework's DAG). If you try to index `Z[1]` or show `Z` at the repl then the computation happens and the result is displayed, and that result will be cached and re-used later. This gives CF a chance to optimise the graph as much as possible while keeping those optimisations as an implementation detail as far as the public API is concerned.

There are some fundamental pros and cons to this approach vs. the macroification above, as well as some things we haven't figured out yet. The upside is that it's even more transparent in principle; all Julia array code works with no changes on `CFArray`s and it usually works *fast* by default. The downside is that the abstraction is potentially leaky; it's not ideal if performance depends heavily on where your computations are reified. `@cf` is completely transparent about what's happening in CF-land and when. Maybe we need both approaches in the long term.

Some issues: (1), intermediates. With the lazy approach we can't tell what parts of the graph may be reused by the programmer in future. We could assume that it won't be reused and re-compute it if we're wrong, but this comes back to the leaky abstraction point. (2) aggregates like `reduce(+, xs)` – if these return numbers then they clearly have to be eager, but if they're eager then multiple aggregates may not be optimal. One solution is return something like a lazy number type that acts as a number but doesn't reify until called, but this seems like a tricky concept to expose to end users.

The solution to these issues may simply be to make them work in an obvious / non-optimal way by default and allow the user to break the abstraction where necessary, even simply by using `@cf` as a fallback.

datnamer:
Does iteration work even on distributed  and  alternatove storage backed arrays?

MikeInnes:
Iteration can certainly be made to work with the lazy CFArray, it probably wouldn't ever be the most efficient approach, but if it implicitly loaded a chunk at a time onto the master process then that would work ok.

The lazy CFArray approach is also nicer for debugging, because if you inspect the array in the middle of a function you see the state of the data, rather than an opaque dag.

MikeInnes:
So the way that ArrayFire solves the intermediates problem is that it uses a C++ language feature to opt in to refcounting (I don't know a lot about this so my terminology may be off). This is the only way I can think of to solve this problem without changing the user-level API. It may be worth considering whether this would be viable as a language-level feature within Julia.

Until then macros will have to be part of the API on some level, and I don't think it's possible to get top performance in a completely transparent way.

MikeInnes:
Closing this in favour of recent changes to the lazy array approach 🎆

shashi:
Ping. #46 might be relevant here. Would like to hear what you think of dropping the expression level information (in order to make the implementation simpler and more readable mainly) given some time has passed since this was last discussed.

PR #37, opened 2016-05-19
timing and profiling

shashi:
See docs in README

This is also going to be the basis of the performance UI. One can create many types of `log_sink` objects (arrays, channels, remotechannels) to aggregate metadata about small chunks of work occurring in parallel. You can collect profiling samples from select events and pretty-print them.

I am trying to understand why this stuff is so slow with profiling turned on.

cc @ViralBShah

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,9 +243,70 @@ $, &, (.!=), (.<), (.<=), (.==), (.>),
 **Upcoming features**
 - Sorting
 
+## Debugging performance
+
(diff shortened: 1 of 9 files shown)


shashi:
@amitmurthy see travis failure on nightly. It's the 0 peer thing we saw.

amitmurthy:
I think I have identified the issue. Will submit a PR tomorrow.

PR #42, opened 2016-08-04
sort

shashi:
Based on http://www.cs.ucsb.edu/~gilbert/reports/sorting111904.pdf

- [x] 0.4 ??
- [x] Respect sort options
- [x] pass context properly


shashi:
@tkelman it seems for a short period of time the release version on travis is going to be 0.4, but the latest is going to be 0.6. BaseTestNext is not available on 0.6 but is required for 0.4, is there any way to deal with this in `test/REQUIRE`?

PR #43, opened 2016-12-15
Some possible fixes

ranjanan:
* Return ComputedArray when loading binary files
* Fix depwarns `isgeneric` -> `isa(f, Function)`

### DIFF
--- a/src/ui/graph.jl
+++ b/src/ui/graph.jl
@@ -2,7 +2,7 @@ import Dagger: Thunk
 export show_plan
 
 function node_label(io, t::Thunk, c)
-    if isgeneric(t.f)
+    if isa(t.f, Function)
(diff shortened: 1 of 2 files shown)


src/ui/graph.jl
```diff
@@ -100,7 +100,7 @@ function show_ast(io, f)
 end
 
 function showfn(io, f::Function)
-    if isgeneric(f)
+    if isa(f, Function)
```
shashi: Well, on 0.4 this would allow you to actually print out the anonymous function.

shashi: This fix makes that feature go away on 0.4, but that's ok. I guess we have to figure out a way to do this on 0.5 then.

PR #62, opened 2017-06-07
delete Domain abstract type, Cats and Views

### LINKED ISSUE #54: Domain abstract type, View and Cat are redundant
With #45 and #46 we should end up with a single array type which is like:

```julia
immutable DaggerArray{T,N} <: AbstractArray{T,N}
    dag::AbstractChunk{<:AbstractArray{T,N}}
end
```

But we can simplify things even further if we have this instead:
```julia
immutable DaggerArray{T,N} <: AbstractArray{T,N}
    size::NTuple{N,Int}
    chunks::Associative{ArrayDomain, Any}
end
```
where `chunks` is an efficient mapping from sub-index-spaces to `Chunk`s and/or `Thunk`s.

To keep things simple we can stop storing `domain` in `Chunk`s. Since the subdomains info has been moved to DaggerArray we can stop tracking it in Cat as well, which makes Cat pretty much an extraneous wrapper equivalent to `Thunk(_hvcat, size(chunks), chunks...)`. By extension `View` also is `Thunk(getindex, chunk ,idxs...)`.

code around Cat, View and Sub were too array-specific anyway and moving this stuff to `DaggerArray` will dump less baggage on non-array implementations on Dagger.

Finally it doesn't make sense for Domain to be an abstract type and `ArrayDomain` to be its subtype. Just drop that relationship and remove Domain. It seems "IndexSpace" is the standard term for this thing according to http://dl.acm.org/citation.cfm?id=898874 so maybe we can use that.

The only thing we will be giving up is having a `View` on a `Cat` be able to turn into a `Cat` of `View`s outside of a direct `getindex` (getindex should be handled efficiently by the newly proposed Array type) which doesn't seem to happen that often in the wild.

shashi:
After a year I can tell you for sure that this was not worth it.

closes #54


src/array/lazy-array.jl
```diff
@@ -18,26 +18,26 @@ function Base.show(io::IO, x::LazyArray)
     @compat show(io, m, x)
 end
 
-immutable ComputedArray{T,N} <: LazyArray{T, N}
+immutable DArray{T,N} <: LazyArray{T, N}
```
andreasnoack: Isn't it a bit confusing with two different `DArray` types named `DArray`?

shashi: Yeah, I think so!

Issue #46, opened 2017-02-13
Get rid of operation-level DAG


shashi:
Since we now have things like [syntactic loop fusion](http://julialang.org/blog/2017/01/moredots) in the language, and in general it is possible to have the benefits of the high level API by syntactic means such as https://github.com/MikeInnes/DataFlow.jl, it will simplify implementation to just use:
```julia
immutable DaggerArray{T,N} <: AbstractArray{T,N}
    dag::AbstractChunk{<:AbstractArray{T,N}}
end
```
As the array type. Operations will simply return the next dag wrapped in an `DaggerArray` object.

This along with #45 should simplify this package drastically.

jpsamaroo:
I'm against this; I think it's better if we start exploiting operation-level DAGs more than we do currently, by formalizing and exporting the staging API and allowing users to provide their own types which can be staged to generate a partial DAG at scheduling time (#173 ).

Issue #47, opened 2017-02-13
Block distribution incompatibility in matmul should fail better


shashi:
```julia
X1 = Distribute(BlockPartition(10, 20), X)
@test_throws DimensionMismatch compute(X1*X1)
```
But throws an alarming BoundsError

jpsamaroo:
This throws `DimensionMismatch` on master as desired, so I'm going to close this issue. Do please re-open if you have another example with fails with an inappropriate error.

PR #69, opened 2017-11-11
WIP: Slice utility + fix map

MikeInnes:
Here's a small utility we need to mix arrays of different block layouts in general. To show it working I've fixed `map` to handle this case, e.g.

```julia
using Dagger

xs = distribute(rand(10,10), Blocks(5,5))
ys = distribute(rand(10,10), Blocks(3,3))

map(+, xs, ys) |> collect
```

There's some overhead added in the case that currently works, not sure what the impact will be but it it's easy to optimise if needed.

In future I think this should just be called `getindex`. That will be possible with #46 and a more general notion of chunks; e.g. if you can pass a `DArray` directly to a thunk.

### DIFF
--- a/src/array/getindex.jl
+++ b/src/array/getindex.jl
@@ -40,3 +40,20 @@ Base.getindex(c::ArrayOp, idx::ArrayDomain) = GetIndex(c, indexes(idx))
 Base.getindex(c::ArrayOp, idx...) = GetIndex(c, idx)
 Base.getindex(c::ArrayOp, idx::Integer...) =
    compute(GetIndexScalar(c, idx))
+
+# Internal utility
+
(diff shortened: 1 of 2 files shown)


jpsamaroo:
@MikeInnes this will need a rebase.

Is this a solution for #47 as well, or are the two unrelated?

MikeInnes:
It definitely seems related and the utilities here could be probably be reused for matmul. It's been a long time since I looked over it though, so it's unlikely I'll have time to fix this up. Feel free to close if no-one else is likely to get to it.

jpsamaroo:
No worries, I'll pick this one up if I can figure it out. Thanks for the quick response :)

PR #73, opened 2018-02-02
BlockIO: fix find_end_pos/find_start_pos in constructor

iblis17:


jpsamaroo:
Hi @iblis17 , can you recall what this PR is intended to fix/improve? Do you have an MWE that this fixes?

iblis17:
Hi @jpsamaroo , there are two commits in this PR.
Seems that the first one has been included in the current master.
The second one is for performance tuning.
But I'm not sure that tuning is still true to current Julia compiler or not.

jpsamaroo:
I see... if you happen to have some example code that we can use to benchmark this change, that'd be awesome. I don't know enough about the BlockIO code to be able to guess what impact this change could have.

iblis17:
well, no benchmark snippets on hand... too long ago.

PR #75, opened 2018-02-21
make the scheduler configurable

shashi:
creating a PR to make this easy to review.

-----------------------------------
updating some details:

This makes the dag scheduler and executor a configurable module, with the existing implementation included as the default.

A scheduler/executor module must implement:
- `compute_dag(ctx, d::Dagger.Thunk)`: schedules the dag, waits for the top node (`d`) and returns its result
- `cleanup(ctx)`: cleans up any state/IPC before process exit

To switch to a different implementation, set the full module name that implements the above methods in the plugin config. E.g:
```
Dagger.PLUGIN_CONFIGS[:scheduler] = "MyCustomModule.MyScheduler"
```

Apart from that this PR also includes:
- Add the thunk or chunk id to the label when visualizing dag with `show_plan`
- Option not to use shared arrays (in array sort).
- Rollover thunk ids, limit them to 30bits. Need to rollover when processes are long running. And keeping the remaining bits for scheduler/executor to pass annotations around easily.
- Disable cache tests till it is reworked on. They fail at the moment (unrelated to this PR).
- Disables affinity propagation in array sort, which was setting specific affinities on dag nodes to get a uniform distribution of the result. Maybe distributing result should be a separate step.
- Avoids calling `compute` from worker processes. Workers are considered executors and master the scheduler.

### DIFF
--- a/src/Dagger.jl
+++ b/src/Dagger.jl
@@ -5,6 +5,11 @@ using Compat
 
 import Base.collect
 
+const PLUGINS = Dict{Symbol,Any}()
+const PLUGIN_CONFIGS = Dict{Symbol,String}(
+    :scheduler => "Dagger.Sch"
+)
+
 include("lib/util.jl")
 include("lib/logging.jl")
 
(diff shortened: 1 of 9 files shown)


tanmaykm:
will merge in a bit

PR #82, opened 2018-03-15
Improve efficiency of pop_with_affinity! when there are many tasks

donm:
These changes remove the biggest inefficiencies in computing a DAG with a large number of tasks. For example, the following benchmark runs in about 6.5-7.0 seconds before the change, and runs in about 1.0 seconds with the PR.

```
addprocs(4)
using Dagger

@everywhere function spin()
    a = 0.0
    for i in 1:5000000
        a = sqrt(i)
    end
    return a
end

function main(n)
    op = delayed(spin)
    leafs = [op() for i in 1:n]
    result = delayed((args...)->nothing)(leafs...)
    compute(result)
end
main(10)
@time main(1000)
```

As a comparison, this workflow implemented without Dagger (using spawned workers that consume a `RemoteChannel`) runs in about 0.7 seconds on my machine.

### DIFF
--- a/src/scheduler.jl
+++ b/src/scheduler.jl
@@ -76,32 +76,41 @@ function compute_dag(ctx, d::Thunk)
 end
 
 function pop_with_affinity!(ctx, tasks, proc, immediate_next)
+    # allow JIT specialization on Pairs
+    mapfirst(c) = first.(c)
+
     if immediate_next
         # fast path
-        if proc in first.(affinity(tasks[end]))
+        if proc in mapfirst(affinity(tasks[end]))
             return pop!(tasks)
         end
     end
-    parent_affinities = affinity.(tasks)
+
+    # TODO: use the size
+    parent_affinity_procs = Vector(length(tasks))
+    # parent_affinity_sizes = Vector(length(tasks))
     for i=length(tasks):-1:1
-        # TODO: use the size
-        if proc in first.(parent_affinities[i])
-            t = tasks[i]
+        t = tasks[i]
+        aff = affinity(t)
+        aff_procs = mapfirst(aff)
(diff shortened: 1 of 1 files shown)


shashi:
That is awesome! It's funny I tried to fix this last night as well! My approach was to try and stop using the `Processor` abstract type...

https://github.com/JuliaParallel/Dagger.jl/commit/64362c512c4bc1920abfe07d1efe0a6cd6dfaf66

Maybe both these will help?

donm:
That is a funny coincidence!  Yeah, I see how changing `Processor` could help.  I'll try to check back later and test the merge of the two commits.

shashi:
Just curious, how did you find out what was causing the slowdown?

shashi:
Not much changes if I apply both:

This patch:
```
1.349310 seconds (3.81 M allocations: 395.845 MiB, 3.00% gc time)
```
Plus my patch:
```
1.297081 seconds (3.27 M allocations: 386.702 MiB, 2.71% gc time) 
```

donm:
When I ran something similar to `@profile main()` (using the snippet above), the times for the `float.` broadcasts just really stood out.

I started looking into it because the first way that I could find to compute a bunch of independent DAGs in parallel was to feed them all into a final dummy node and compute that.  But is there another way that I haven't stumbled across yet?

shashi:
You could use `Dagger.treereduce(delayed(vcat), tasks)` -- many dummy nodes, but gives better parallelism.

I guess in this case figuring out the problem was easy because it's on the master process -- it's harder if something is slow on the workers and that's Dagger's fault. Hoping https://github.com/JuliaParallel/DProfile.jl may help with this.

PR #91, opened 2018-06-20
Remove invalid export statements

ScottPJones:


ScottPJones:
Bump: the only failure is from issues with nightlies and 32-bit Windows that have nothing to do with this PR,(and now that I fixed the appveyor.yml, it now passes on 64-bit Windows on v0.6.x of Julia)

PR #93, opened 2018-08-24
Remove Logging from REQUIRE

andreasnoack:
It looks like it isn't used anymore and I noticed that it can cause some trouble now that we also have a stdlib with the same name.


JeffBezanson:
Thanks. This is now part of #92.

PR #104, opened 2019-04-04
 Don't set finalizers for DRef chunks as they are already set in MemPool

nkottary:


shashi:
Actually can't we just add this to serialize deserialize of DRef?? I think that's more correct, we can just delete any finalizer code on Chunks

nkottary:
I've added `deserialize` to MemPool: https://github.com/JuliaComputing/MemPool.jl/pull/28

but I get a `StackOverflowError` now.

shashi:
What's a minimal example?

shashi:
fixed it

jpsamaroo:
Closing since the DRef is now refcounted.

PR #106, opened 2019-06-06
Lowerbound to Julia 1.0 and add Project.toml

jpsamaroo:
Added Project.toml and excluded Manifest.toml in gitignore
Bumped Julia version to 1.0 in REQUIRE

Hopefully the `[compat]` line is correct too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.jl.mem
 .*~
 .*sw?
+Manifest.toml
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,24 @@
+name = "Dagger"
+uuid = "d58978e5-989f-55fb-8d15-ea34adc7bf54"
+version = "0.8.0"
+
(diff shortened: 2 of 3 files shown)


jpsamaroo:
Any objections to what I've done here? We'll need a Project.toml if we're to register new versions in General, and this makes working with a `dev`'d Dagger project much easier to work with. So I'd like to get this PR in as soon as possible.

aviks:
LGTM

jpsamaroo:
Thanks! Would now be a good time to tag a minor/major release of Dagger?

andreasnoack:
Yes. Please check that JuliaDB still works with these changes before making the release.

jpsamaroo:
I already did a local run of the JuliaDB tests on Dagger master about a month ago, but I would like at least 1 more person to test that as well, just in case. I was hoping to recruit such a soul during the JuliaCon hackathon :)

PR #111, opened 2019-07-26
Documentation improvement

abx78:
documentation improvement

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,39 +6,81 @@
 
 At the core of Dagger.jl is a scheduler heavily inspired by [Dask](http://dask.pydata.org/en/latest/). It can run computations represented as [directed-acyclic-graphs](https://en.wikipedia.org/wiki/Directed_acyclic_graph) (DAGs) efficiently on many Julia worker processes.
 
-# DAG creation interface
 
-Here is an example DAG:
+## Installation
+
+You can install Dagger by typing
+
+```julia
+julia> ] add Dagger
+```
+
+## Usage
+
+Once installed, the `Dagger` package can by used by typing
 
 ```julia
 using Dagger
-
-p = delayed(f; options...)(42)
-q = delayed(g)(p)
-r = delayed(h)(53)
-s = delayed(combine)(p, q, r)
 ```
-The connections between nodes `p`, `q`, `r` and `s` is represented by this dependency graph:
 
-![graph](https://user-images.githubusercontent.com/25916/26920104-7b9b5fa4-4c55-11e7-97fb-fe5b9e73cae6.png)
+The main function for using Dagger is *delayed*
 
 `delayed(f; options...)`
 
-Returns a function which when called creates a `Thunk` object representing a call to function `f` with the given arguments. If it is called with other thunks as input, then they form a graph with input nodes directed at the output. The function `f` get the result of evaluating the input thunks.
-
-To compute and fetch the result of a thunk (say `s`), you can call `collect(s)`. `collect` will fetch the result of the computation to the master process. Alternatively, if you want to compute but not fetch the result you can call `compute` on the thunk. This will return a `Chunk` object which references the result. If you pass in a `Chunk` objects as an input to a delayed function, then the function will get executed with the value of the `Chunk` -- this evaluation will likely happen where the input chunks are, to reduce communication.
- 
-Options to `delayed` are:
+Returns a function which when called creates a `Thunk` object representing a call to function `f` with the given arguments. If it is called with other thunks as input, then they form a graph with input nodes directed at the output. The function `f` get the result of the input Thunks.
+Thunks don't pass keyword argument to the function `f`. Options kwargs... to `delayed` are passed to the scheduler to control its behavior:
 - `get_result::Bool` -- return the actual result to the scheduler instead of `Chunk` objects. Used when `f` explicitly constructs a Chunk or when return value is small (e.g. in case of reduce)
 - `meta::Bool` -- pass the input “Chunk” objects themselves to `f` and not the value contained in them - this is always run on the master process
 - `persist::Bool` -- the result of this Thunk should not be released after it becomes unused in the DAG
 - `cache::Bool` -- cache the result of this Thunk such that if the thunk is evaluated again, one can just reuse the cached value. If it’s been removed from cache, recompute the value.
 
+# DAG creation interface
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,34 +6,81 @@
...
+add1(value) = value+1
+add2(value) = value+2
+combine(a...) = sum(a)
+
+p = delayed(add1)(4)
+q = delayed(add2)(p)
+r = delayed(add1)(3)
 s = delayed(combine)(p, q, r)
+
+collect(s)
```
jpsamaroo: For clarity, maybe make this `@assert collect(s) == 16` so that it's clear that `collect` returns the final result (and also illustrates how `delayed` is just a set of "regular" function calls).

README.md
```diff
@@ -6,34 +6,81 @@
...
+### Polytree
+
+[Polytrees](https://en.wikipedia.org/wiki/Polytree "Polytrees") are easily supported by Dagger. To do the trick, pass all the head nodes `Thunks` in a `Thunk` that you will create as a top node for the graph end compute it.
+```
+group(x...) = [x...]
+top_node = delayed(group)(head_nodes...)
+compute(top_node)
+```
+
+# DAG creation interface
```
jpsamaroo: I think this could stand to move back up to before the examples, since the first section after this isn't about DAG creation but about scheduling.

README.md
```diff
@@ -6,34 +6,81 @@
...
 Options to `delayed` are:
 - `get_result::Bool` -- return the actual result to the scheduler instead of `Chunk` objects. Used when `f` explicitly constructs a Chunk or when return value is small (e.g. in case of reduce)
 - `meta::Bool` -- pass the input “Chunk” objects themselves to `f` and not the value contained in them - this is always run on the master process
 - `persist::Bool` -- the result of this Thunk should not be released after it becomes unused in the DAG
 - `cache::Bool` -- cache the result of this Thunk such that if the thunk is evaluated again, one can just reuse the cached value. If it’s been removed from cache, recompute the value.
 
+
+### Polytree
+
+[Polytrees](https://en.wikipedia.org/wiki/Polytree "Polytrees") are easily supported by Dagger. To do the trick, pass all the head nodes `Thunks` in a `Thunk` that you will create as a top node for the graph end compute it.
```
jpsamaroo: "To do the trick" -> "To make this work with Dagger"

README.md
```diff
@@ -6,34 +6,81 @@
...
+
+```julia
+using Dagger
+```
+
+The main function for using Dagger is *delayed*
+
+`delayed(f; options...)`
+
+When called creates a `Thunk` object representing a call to function `f` with the given arguments. If it is called with other thunks as input, then they form a graph with input nodes directed at the output. The function `f` get the result of the input Thunks.
```
jpsamaroo: Would you mind adding a comment that `Thunk`s don't pass keyword argument to the function `f`, and that the `options...` kwargs are passed to the scheduler to control its behavior?

jpsamaroo: One other thing - why was this changed to remove "Returns a function which..."? I'm pretty this is now less accurate.

README.md
```diff
@@ -6,34 +6,80 @@
...
+Thunks don't pass keyword argument to the function `f`.
 
-To compute and fetch the result of a thunk (say `s`), you can call `collect(s)`. `collect` will fetch the result of the computation to the master process. Alternatively, if you want to compute but not fetch the result you can call `compute` on the thunk. This will return a `Chunk` object which references the result. If you pass in a `Chunk` objects as an input to a delayed function, then the function will get executed with the value of the `Chunk` -- this evaluation will likely happen where the input chunks are, to reduce communication.
- 
-Options to `delayed` are:
+
+Options kwargs... to `delayed` are passed to the scheduler to control its behavior:
```
jpsamaroo: Hmm, ok I see that you actually put this down here... I would prefer that this goes inline with "Thunks don't pass keyword argument to the function `f`.".

README.md
```diff
@@ -6,34 +6,80 @@
...
+The final result is the obvious consequence of the operation
+
+ `add1(4)` + `add2(add1(4))` + `add1(3)`
+
+ `(4 + 1)` + `((4 + 1) + 2)` + `(3 + 1)` = 16
+
+To compute and fetch the result of a thunk (say `s`), you can call `collect(s)`. `collect` will fetch the result of the computation to the master process. Alternatively, if you want to compute but not fetch the result you can call `compute` on the thunk. This will return a `Chunk` object which references the result. If you pass in a `Chunk` objects as an input to a delayed function, then the function will get executed with the value of the `Chunk` -- this evaluation will likely happen where the input chunks are, to reduce communication.
+
+The key point is that, for each argument to a node, if the argument is a `Thunk`, it'll be executed before this node and its result will be passed into the function `f` provided.
+If the argument is *not* a `Thunk` (just some regular Julia object), it'll be passed as-is to the function `f`
```
jpsamaroo: Nitpick: missing a period.

README.md
```diff
@@ -6,34 +6,80 @@
...
+
+### Polytree
+
+[Polytrees](https://en.wikipedia.org/wiki/Polytree "Polytrees") are easily supported by Dagger. To make this work, pass all the head nodes `Thunks` in a `Thunk` that you will create as a top node for the graph end compute it.
```
jpsamaroo: Nitpicks:

"all the head nodes `Thunks`" -> "all the head node `Thunk`s"
"in a `Thunk` that you will create as a top node for the graph end compute it." - This sounds awkward to me, maybe something like: "into a call to `delayed` as arguments, which will act as the top node for the graph."

README.md
jpsamaroo: This is still down here in the diff?

README.md
```diff
@@ -6,34 +6,80 @@
 
 At the core of Dagger.jl is a scheduler heavily inspired by [Dask](http://dask.pydata.org/en/latest/). It can run computations represented as [directed-acyclic-graphs](https://en.wikipedia.org/wiki/Directed_acyclic_graph) (DAGs) efficiently on many Julia worker processes.
 
-# DAG creation interface
 
-Here is an example DAG:
+## Installation
+
+You can install it by typing
```
jpsamaroo: "You can install it" -> "You can install Dagger"

jpsamaroo:
Thanks @abx78 !

PR #112, opened 2019-07-26
Add some scheduler documentation

jpsamaroo:
Including fault tolerance docs that I should've added previously...

### DIFF
--- a/src/fault-handler.jl
+++ b/src/fault-handler.jl
@@ -6,9 +6,27 @@ check_exited_exception(res::RemoteException) =
 check_exited_exception(res::ProcessExitedException) = true
 check_exited_exception(res) = false
 
+"""
+    handle_fault(...)
+
+An internal function to handle a worker dying or being killed by the OS.
+Attempts to determine which `Thunk`s require rescheduling based on a
+"deadlist", and then corrects the scheduler's internal `ComputeState` struct
+to recover from the fault.
+
+Note: The logic for this functionality is not currently perfectly robust to
+all failure modes, and is only really intended as a last-ditch attempt to
+repair and continue executing. While it should never cause incorrect execution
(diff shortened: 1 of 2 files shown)


src/fault-handler.jl
```diff
@@ -6,9 +6,27 @@ check_exited_exception(res::RemoteException) =
 check_exited_exception(res::ProcessExitedException) = true
 check_exited_exception(res) = false
 
+"""
+    handle_fault(...)
+
+An internal function to handle a worker dieing or being killed by the OS.
```
abx78: you might want to use `dying` here as more common rather than `dieing` (very specialized term)

jpsamaroo: Good catch :slightly_smiling_face:

abx78: My first review is about dead stuff 💀 sob 🗡

jpsamaroo:
Any Dagger maintainers care to review this? If not I'll merge in a few days.

src/fault-handler.jl
```diff
@@ -6,9 +6,27 @@ check_exited_exception(res::RemoteException) =
...
+to recover from the fault.
+
+Note: The logic for this functionality is not currently perfectly robust to
+all failure modes, and is only really intended as a last-ditch attempt to
+repair and continue executing. While it should never cause incorrect execution
+of DAGs, it *may* cause a `KeyError` or other failures in the scheduler due to
+the complexity of getting the internal state back to a consistent and proper
+state.
+"""
 function handle_fault(ctx, state, thunk, oldproc, chan, node_order)
```
abx78: Is it maybe worth to mention what the arguments of the handle_func are?

jpsamaroo: I would, if only I knew the purpose of each argument :smile:

PR #113, opened 2019-09-05
Threading trick!

jpsamaroo:
1. Spawn `Threads.@spawn f(x...)` instead of `f(x...)` where `x` is the "fetched" data.
2. repeat every process as many times as nthreads() -- causing Dagger to oversubscribe each worker processors as many times as the threads. TOOD: actually get nthreads from each worker rather than using the value on the master process

By @shashi 

Will need to be adjusted so that pre-MT julia still works, and to make this an opt-in feature for now.

### DIFF
--- a/src/processor.jl
+++ b/src/processor.jl
@@ -17,10 +17,11 @@ mutable struct Context
     procs::Array{OSProc}
     log_sink::Any
     profile::Bool
+    options
 end
 
 function Context(xs)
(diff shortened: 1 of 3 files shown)


jpsamaroo:
The nightly error is somehow related to Dagger not being run with threads; I'll have to figure out how to enable them in Travis, or failing that, disable or neuter the threading tests...

jpsamaroo:
Thanks @shashi and @tanmaykm !

PR #116, opened 2020-02-04
Remove usage of Base.@get!

jpsamaroo:


jpsamaroo:
MacOS test failure is unrelated.

PR #117, opened 2020-02-04
Add dynamic scheduler control

jpsamaroo:
Allow thunks to dynamically control the scheduler

Todo:
- [x] `exec!`
- [x] `add_thunk!`
- [x] `get_dag_ids`
- [x] Change API to use `task_local_storage` for scheduler handle
- [x] Export dynamic API from Sch
- [x] Test `add_thunk!` and `wait`/`fetch!` better
- [x] Update docs for TLS API
- [ ] ~Expose scheduler control through `Context` somehow~ For a future PR
- [x] Simplify RPC interface to just use `exec!`
- [x] Change `set_return!` to `wait`/`fetch`, to allow one thunk to wait on another
- [x] Use special ThunkID instead of special Arg

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Dagger"
 uuid = "d58978e5-989f-55fb-8d15-ea34adc7bf54"
-version = "0.10.1"
+version = "0.10.2"
 
 [deps]
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -16,6 +16,7 @@ makedocs(;
         "Processors" => "processors.md",
         "Scheduler Internals" => "scheduler-internals.md",
         "Logging and Graphing" => "logging.md",
+        "Dynamic Scheduler Control" => "dynamic.md",
     ]
 )
 
--- a/docs/src/dynamic.md
+++ b/docs/src/dynamic.md
@@ -1,0 +1,54 @@
+# Dynamic Scheduler Control
+
+Normally, Dagger executes static graphs defined with `delayed` and `@par`.
+However, it is possible for thunks to dynamically modify the graph at runtime,
+and to generally exert direct control over the scheduler's internal state. The
+`Dagger.sch_handle` function provides this functionality within a thunk:
+
+```julia
+function mythunk(x)
+    h = Dagger.sch_handle()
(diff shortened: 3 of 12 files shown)


jpsamaroo:
@shashi this is the foundation of what I was discussing with you in Slack today.

Basically, we can expand this code to support modifying the DAG at execution time, with the scheduler support, of course.

jpsamaroo:
Bump. Any objections to the idea or implementation?

src/scheduler.jl
```diff
@@ -82,6 +112,10 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
 
     node_order = x -> -get(ord, x, 0)
     state = start_state(deps, node_order)
+    # setup worker-to-scheduler channels
+    for p in ps
+        state.worker_chans[p.pid] = fetch(@spawnat p.pid RemoteChannel(1))
```
vchuravy:
```suggestion
        state.worker_chans[p.pid] = remotecall_fetch(RemoteChannel, p.pid,1)
```

jpsamaroo:
Thanks for the review! I'll address your recommendations once I come back to this PR, and I also think I'll hold off on merging this until we expose a few more controls through this API, so that we can see what it will ultimately look like. I'd like to at least support adding new thunks to the DAG, and querying the scheduler state and the full DAG structure from this interface.

src/scheduler.jl
jpsamaroo: Addressed in latest push, although the suggestion doesn't work except on recent Julia versions, because `remotecall_fetch` didn't previously accept constructors as the first argument.

jpsamaroo:
Note to onlookers: locking was added to allow users to add their own `process_dynamic!` methods which interact with the scheduler state without stepping on the scheduler's proverbial toes. Finer-grained locking will probably be necessary later on to enable more efficient scheduler modifications. Providing a more fine-grained semi-internal DAG modification API should make concurrent modification much easier and more efficient.

jpsamaroo:
I don't think I'm going to keep the `dynamic` kwarg; it's ugly and makes the thunk API ugly. We should be able to instead pass the scheduler handle through `task_local_storage`, which will let us keep thunks working as normal. The only price we pay is getting the scheduler handle each time, which should only be the cost of a dynamic dispatch; this should be insignificant given that most operations on the handle involve network traffic.

PR #119, opened 2020-02-09
Housekeeping

visr:


jpsamaroo:
LGTM. Looks like the Travis Linux failures are due to the following `try` block not having a `catch` or `finally`: https://github.com/JuliaParallel/Dagger.jl/blob/5266b4fefc889f899cf36ef142864bccc415a953/src/lib/block-io.jl#L14-L16

If you wouldn't mind fixing this (just add an empty `finally` block) so we can get Travis Linux passing, then we can merge this.

jpsamaroo:
Thanks!

PR #120, opened 2020-02-10
Polish up docstrings and more

visr:
Sorry if this is too much of a mixed bag, would be happy to split it up as well. But it's mostly straightforward.

88d9964ca78a18d2c42e9a134184a588dbaee81e removes empty files
d6aca842107ff2a4fedc865a28daa31c2425a7d0 only touches docstrings
4cd00516c18c1cf2d2953e4e7cdee9d1f190d94c adds argument type annotations, good to double check if all are correct
82c15067199640f9895994fea3e8b8c538cdd89e is a simple do block
a4c409b7737c088250789ec228eb292c12f1e9d2 DArray.concat must be a function right?
d8f4fdee5b10a3cb58e03db2dacfd74dd88b10a2 looks like an oversight

### DIFF
--- a/src/array/darray.jl
+++ b/src/array/darray.jl
@@ -39,12 +39,15 @@ function getindex(a::ArrayDomain, b::ArrayDomain)
 end
 
 """
-    alignfirst(a)
+    alignfirst(a) -> ArrayDomain
 
-Make a subdomain a standalone domain. For example,
+Make a subdomain a standalone domain.
 
-    alignfirst(ArrayDomain(11:25, 21:100))
-    # => ArrayDomain((1:15), (1:80))
+# Example
+```julia-repl
+julia> alignfirst(ArrayDomain(11:25, 21:100))
+ArrayDomain((1:15), (1:80))
+```
 """
 alignfirst(a::ArrayDomain) =
     ArrayDomain(map(r->1:length(r), indexes(a)))
@@ -59,7 +62,11 @@ ndims(a::ArrayDomain) = length(size(a))
 isempty(a::ArrayDomain) = length(a) == 0
 
 
-"The domain of an array is a ArrayDomain"
+"""
+    domain(x::AbstractArray) -> ArrayDomain
+
+The domain of an array is an ArrayDomain.
+"""
 domain(x::AbstractArray) = ArrayDomain([1:l for l in size(x)])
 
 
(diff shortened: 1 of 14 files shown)


src/compute.jl
```diff
@@ -84,10 +91,12 @@ function dependents(node::Thunk, deps=Dict{Thunk, Set{Thunk}}())
 end
 
 """
-recursively find the number of taks dependent on each task in the DAG.
-Input: dependents dict
+    noffspring(dpents::Dict{Thunk, Set{Thunk}}) -> Dict{Thunk, Int}
+
+Recursively find the number of taks dependent on each task in the DAG.
```
jpsamaroo: Typo (from original source): `taks` -> `tasks`

src/fault-handler.jl
```diff
@@ -1,4 +1,8 @@
-"Recursively checks if an exception was caused by a worker exiting."
+"""
+    check_exited_exception(res) -> Bool
```
jpsamaroo: This is technically an internal function only used by the fault handler, but I think this is still a fine change. Maybe just add `res::Exception` to the argument to indicate that it takes an `Exception` as input?

src/processor.jl
```diff
@@ -3,31 +3,45 @@ export OSProc, Context
...
+OS process - contains pid returned by `Distributed.workers`
 """
 struct OSProc <: Processor
     pid::Int
 end
 
 """
-A context represents a set of processors to use
-for a papply operation.
+A context represents a set of processors to use for a operation.
```
jpsamaroo: `a operation` -> `an operation`

src/processor.jl
```diff
@@ -3,31 +3,45 @@ export OSProc, Context
...
+    Context(xs::Vector{OSProc}) -> Context
+    Context(xs::Vector{Int}) -> Context
+
+Create a Context, by default adding each available worker once from
+every available thread. Use the `nthreads` keyword to use a different
+number of threads.
+
+It is also possible to create a Context from a vector of [`OSProc`](@ref),
+or equivalently the underlying process ids can also be passed directly
+as a Vector{Int}.
```
jpsamaroo: Should `Vector{Int}` be between backticks?

visr:
Thanks for having a look! I addressed the comments.

src/array/darray.jl
```diff
@@ -237,7 +237,7 @@ end
 A DArray object may contain a thunk in it, in which case
 we first turn it into a Thunk object and then compute it.
 """
-function compute(ctx, x::DArray; persist=true, options=nothing)
+function compute(ctx::Context, x::DArray; persist=true, options=nothing)
```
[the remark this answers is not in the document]
visr: I cannot answer this, but just want to mention that a part of the signatures already had `::Context`. So probably we should either remove them all or add them all.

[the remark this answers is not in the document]
jpsamaroo: Yeah I agree with you, I guess we can just go with your changes, and then widen things as needed later on.

jpsamaroo:
Thanks a ton!

PR #121, opened 2020-03-31
Support scheduling to other Processor types

jpsamaroo:
Added PROCESSOR_CALLBACKS for registering children processors
Added convert_arg and execute! methods to control execution
Added proctypes field to Sch.ThunkOptions
Modified scheduler naively to use children processors

Todo:
- [x] Add tests using a fake `Processor`
- [x] Be lazy about moving results back to `from_proc` in `do_task()` when possible (add processor field to `Chunk`)
- [x] Cache `OSProc` on each node
- [x] Expand `iscompatible` to also dispatch on thunk function (falls back to `true`)
- [x] Implement consistent opt-in/opt-out of processor usage, and by default opt-in to CPU and opt-out of GPU processors
- [x] Ensure generic move operates remotely as-needed
- [ ] Run benchmarks which show this approach being approx. equal or faster than before with threads (optionally: with GPUs)
- [x] Add documentation
- [x] Add GPU integration package DaggerGPU.jl to docs


jpsamaroo:
Ok I think this is ready for review. Benchmark numbers (last I checked) were on-par or slightly better than before, and 2X faster with GPUs on certain benchmarks. I'm going to put together a proper presentable benchmark and post it here soon.

jpsamaroo:
One thing to note about this PR is that I did not implement any special scheduling rules for the children processors of an `OSProc`. I did this partly because I'm not sure of a good general approach for scheduling in a hierarchical configuration; however I also believe that it will generally become a bottleneck to keep all of the scheduling decisions on the single master node. For future work, I intend to allow the master scheduler to over-subscribe workers with work, and allow worker nodes to perform their own scheduling decisions based on their available resources. This should help Dagger's `Sch` scheduler to work on highly parallel architectures, where a single scheduler can easily become a bottleneck.

src/processor.jl
```diff
@@ -1,19 +1,237 @@
...
+end
+const OSPROC_CACHE = Dict{Int,OSProc}()
+OSProc(pid::Int=myid()) = get!(OSPROC_CACHE, pid) do
+    remotecall_fetch(get_osproc, pid, pid)
+end
+function get_osproc(pid::Int)
+    proc = OSProc(pid, Dict{Symbol,Any}(), Processor[], Processor[])
+    for cb in PROCESSOR_CALLBACKS
+        try
+            child = cb(proc)
```
vchuravy, line 144: `invokelatest`

jpsamaroo: Good catch! Fixed and added a test for this :smile:

jpsamaroo:
Failures are unrelated to this PR (partially fixed by #122)

PR #123, opened 2020-04-17
added a processor subtype to run on KernelAbstractions

dnabanita7:
I have used OSProc as the parent which is a CPU (OS) process. I have only added a file for DaggerKA.jl package. I am creating another package named DaggerKA.jl and will be sending you the link. 

- [x] Add tests to check
- [x] Ask for any more edits needs to be done
- [x] Added tests for addition of vectors on CPU.
- [ ] Setup GPU CI


jpsamaroo:
This is a good start. The `src/DaggerKA.jl` file is incorrect; you shouldn't make a copy of `src/Dagger.jl`, even if you were splitting this out into a package. For now, you can remove `src/DaggerKA.jl` and start adding tests to `test/processors.jl` that test `DaggerKAProc`.

jpsamaroo:
Ok great, now we'll want a test that creates a KA kernel and launches it. It can be something really simple like an elementwise addition between 2 vectors. Once it works on the CPU, I can test it locally on a GPU, and we can consider setting up GPU CI.

jpsamaroo:
Maybe I should have been more specific: we want the `DaggerKAProc` to be the thing that executes the kernel. So you construct your `@kernel`, pass that kernel (and arguments) to Dagger via `delayed`, and then run the DAG. The `DaggerKAProc` should then execute the kernel with the provided arguments automatically.

dnabanita7:
Yes, I was confused on that. Ok, I am trying.

test/processors.jl
```diff
@@ -40,23 +48,54 @@ end
...
+    value(a) = a
+    @kernel function add(A, B, C)
+        I = @index(Global)
+        C[I] = A[I] + B[I]
+    end
+    A = delayed(value)(ones(512,512))
+    B = delayed(value)(3*ones(512,512))
+    C = delayed(value)(zeros(512,512))
+    S = delayed(add)(A,B,C)
+    @test all(collect(S) .== 4.0)
```
jpsamaroo, line 71: We should be testing the value of `C` as well; the fact that `S` has a value is somewhat irrelevant, since the point of the kernel is to do an elementwise sum of `A .+ B` and store it in `C`.

test/processors.jl
```diff
@@ -40,23 +48,54 @@ end
...
+            @test Dagger.iscompatible_func(proc, opts, unknown_func)
+            @test Dagger.iscompatible_arg(proc, opts, us)
+            @test Dagger.iscompatible(proc, opts, unknown_func, us, 1)
+        end
+    value(a) = a
+    @kernel function add(A, B, C)
+        I = @index(Global)
+        C[I] = A[I] + B[I]
+    end
+    A = delayed(value)(ones(512,512))
```
jpsamaroo, line 67: We need to be settings the `options` kwarg to `delayed` here to a `ThunkOptions` struct where `proctypes=[DaggerKAProc]`, otherwise these kernels might execute on the wrong kind of processor (specifically, a `ThreadProc`, which won't call `fetch').

src/processor.jl
```diff
@@ -227,6 +227,24 @@ end
...
+struct DaggerKAProc <: Processor
+    owner::Int
+end
+
+iscompatible(proc::DaggerKAProc, opts, f, args...) = true
+iscompatible_func(proc::DaggerKAProc, opts, f) = true
+iscompatible_arg(proc::DaggerKAProc, opts, x) = true
+move(ctx, from_proc::OSProc, to_proc::DaggerKAProc, x) = x
+move(ctx, from_proc::DaggerKAProc, to_proc::OSProc, x) = x
+execute!(proc::DaggerKAProc, f, args...) = fetch(@kernel f(args...))
```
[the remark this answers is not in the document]
dnabanita7: I am sorry for replying late. So, since the processor won't accept a kernel object I passed the kernel function. ``delayed`` accepts only functions so, I omitted the creation of kernel object and rather passed in kernel function. The object which will be created from ``delayed`` will be a Dagger object.

test/processors.jl
```diff
@@ -40,23 +48,54 @@ end
...
+    B = delayed(value)(3*ones(512,512))
+    C = delayed(value)(zeros(512,512))
+    S = delayed(add)(A,B,C)
```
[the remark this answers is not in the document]
dnabanita7: Yes, I had added that earlier but that will create a kernel object for the event. 
``delayed`` passes only functions. We need kernel functions so that dagger can create an object of itself from the kernel function.

test/processors.jl
dnabanita7: Yes, I had tried that but since the kernel function won't be accessible if all the variables are not initialised. So, if I am going to pass values of (A,B,C) to the kernel function, it needs to be stored somewhere. So, the ``S`` object stores the value of ``C``.

test/processors.jl
dnabanita7: ok

jpsamaroo:
For the issue of Dagger's `delayed` not accepting non-functions, I would try working around that by either:

1. Modifying `delayed` so that if it's provided a non-`Function` argument `f`, it wraps `f` in an anonymous function, like: `(args...,) -> f(args...,)`.
2. Modify the `f` field in the `Thunk` struct to accept `Any`. I'm not sure if this will cause performance issues due to dynamic dispatch, but I assume probably not.

src/processor.jl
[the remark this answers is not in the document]
jpsamaroo: @vchuravy is indicating that this isn't valid since `@kernel` generates a kernel with the kernel language, but doesn't actually execute the kernel, or wait on it. What we want here is for `f` to be the function that executes the already-constructed kernel object (see my comment about fixing `delayed` not accepting non-functions), and have that be called and then waited on in `execute!`.

This way, the user writes their kernel with `@kernel` outside of Dagger, and then passes it to Dagger to execute with arguments, and then be waited on by Dagger's scheduler.

test/processors.jl
jpsamaroo: Try making these modifications to your example:
```
function addk(p,q,r)
  r .= p .+ q # Use broadcast to assign into the array `r`, instead of reassigning `r`
end

R = zeros(512,512)
r = delayed(value)(R)
```

And then:
```
collect(j) # We don't care about the return value of `j`
all(R .== 4.0)
```

src/processor.jl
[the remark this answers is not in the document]
dnabanita7: ok, I get it now.

test/processors.jl
dnabanita7: ok

dnabanita7: its `` all(r .== 4.0)``

jpsamaroo:
Note to self about how best to finish implementing this PR: along with having a `Processor` subtype for KA, we should also provide a function like `delayed` that accepts the work dimensions for the kernel, and then constructs a set of parallel `delayed` calls and one or more reducers. That would let Dagger parallelize the crap out of really large kernel invocations, which is really what we want to have happen.

jpsamaroo:
Thanks for all the work, but I've decided to move the relevant functionality in a different form into DaggerGPU.jl

PR #141, opened 2020-08-26
Added worker pressure monitoring

jpsamaroo:
The intent of this PR is to inform the scheduler about the "pressure" on each worker, where pressure is some measure of the ability for a worker to compute a thunk in a timely manner. The scheduler is modified to take this pressure into account every time it tries to schedule a thunk, with the scheduler choosing to schedule ready thunks onto the workers with the least amount of relative pressure. We also allow more than just `Nprocs` thunks to be scheduled at a time; instead, we schedule all thunks marked ready. In the future, I intend to expand this approach to schedule full sub-graphs onto each worker to improve data locality.

Currently the tests are failing for various array operations. Whether we end up removing the array operations from this package or not (@shashi), I'd still like to hear if anyone has any ideas on why this is happening, and how to fix it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,7 +16,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
-MemPool = "0.3"
+MemPool = "0.3.2"
 StatsBase = "0.28, 0.29, 0.30, 0.31, 0.32, 0.33"
 julia = "1.0"
 
--- a/src/array/alloc.jl
+++ b/src/array/alloc.jl
@@ -61,7 +61,8 @@ Base.ones(p::Blocks, dims::Integer...) = ones(p, Float64, dims)
 Base.ones(p::Blocks, dims::Tuple) = ones(p, Float64, dims)
 
 function Base.zeros(p::Blocks, eltype::Type, dims)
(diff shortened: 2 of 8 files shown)


jpsamaroo:
I suspect the issue here is the finalizer at https://github.com/JuliaParallel/Dagger.jl/blob/60cf59ff646744e780315adcb32be37f6da35a9d/src/chunks.jl#L110
I added some tracing locally to figure out which calls caused a `pooldelete` before future `poolget`s on the same `DRef` are issued, which pointed to that finalizer.

It appears that https://github.com/JuliaComputing/MemPool.jl/pull/28 might be the solution to this. @shashi @nkottary does this all sound reasonable?

jpsamaroo:
The latest commit adds support for modifying the internal scheduling function, where the default is just the worker pressure scheduler. I've also added a simple "analysis pass" which can be used to find a linear section of thunks. The next step is to run this pass at the beginning of each scheduling step and schedule all of the returned thunks onto the same worker simultaneously, without any unnecessary network traffic. More analyses like these can be added for other common patterns, which (in the right circumstances) will reduce network traffic and underworked nodes.

jpsamaroo:
I've modified the scheduler to return the new pressure after each thunk returns, which is calculated based on the total number of thunks executing on a given processor within a given cluster. This should at least ensure that concurrently executing schedulers play nicely together. In the future, I plan to add runtime data collection from the OS which will give us hard numbers on actual utilization so that shared servers don't get oversubscribed.

jpsamaroo:
CI is failing due to pre-existing DRef issues, but is passing locally for me. I'm going to merge this for now so people can test with it, but I plan to fix the DRef issues before making a new release.

PR #146, opened 2020-10-01
Don't serialize ctx or proc to workers

jpsamaroo:
Per discussion with @DrChainsaw, there's no good reason to serialize the context or processor to another node, and doing so is becoming increasingly more expensive.


shashi:
Nice!

PR #147, opened 2020-10-01
Add functionality to add and remove workers while scheduler is running

### LINKED ISSUE #145: Push new workers to existing Context?
From my understanding, the Context basically plays the role of a WorkerPool. 

Question is if it should be allowed to push new workers to a Context while the scheduler is working on the DAG?

It does not seem to work as is right now as the condition for popping and firing new tasks does not look for new workers in the Context. When adding this to the compute_dag its seems like new workers are being used, but perhaps there are more subtle issues with this approach? This obviously requires fixing #143 as well.

If not I can make a PR for adding this to the existing scheduler.

DrChainsaw:
Fixes #145 

I went for the solution centered around procs as the merits of callbacks is not certain. Putting the launching of new workers in a separate method also allows for people to extend (by using a different context) if they really need to. Maybe this is a bit overengineered. It is a bit annoying to have to setdiff in each iteration but due to the possibility of rmprocs its hard to avoid it.

The option to provide a single worker through `options` makes it awkward to use `procs(ctx)` in the scheduling loop so for now `ps` is still being used. The option of single worker is also handled through a number of `ps === procs(ctx)` checks which is maybe not so nice. Why would anyone use it over creating a context with just one proc in it?

As per @jpsamaroo comments in #145, I didn’t make any callback when a worker is available and the only promise made in rmprocs! docstring is that no new tasks will be scheduled for the process. 

The faulthandler is currently a bit scary as it just plucks a random worker. Shouldn’t it pick one from the context? What if the picked worker is not setup with everywhere yet? This took me longer than I’d like to admit to figure out as the shall_remove_proc function did not play well with this at all: The whole thing just stalled when running the fault-tolerance tests since new workers where not in the context so no tasks were scheduled for them. :)

New Context constructor is to avoid semver breakage as the all args constructor is part of the public API.

I can add a section to the docs as well as soon as the implementation has converged. 

Shall I make addprocs! part of the public API? 
Should it be called addprocs to align with Distributed?
Same as above for rmprocs!.

I don’t love how complicated the testcases for this are. Any suggestions on how to simplify them?

### DIFF
--- a/src/processor.jl
+++ b/src/processor.jl
@@ -284,13 +284,27 @@ Base.lock(f, ctx::Context) = lock(f, ctx.proc_lock)
 """
     addprocs!(ctx::Context, xs)
 
-Add new workers `xs` to an existing `Context ctx`. 
+Add new workers `xs` to `ctx`. 
 
-Workers will typically be assigned new tasks in the next scheduling iteration.
+Workers will typically be assigned new tasks in the next scheduling iteration if scheduling is ongoing.
 
-Workers can be either `Processors` or the underlying process ids as `Integer`s.
+Workers can be either `Processor`s or the underlying process ids as `Integer`s.
 """
 addprocs!(ctx::Context, xs::AbstractVector{<:Integer}) = addprocs!(ctx, map(OSProc, xs))
 addprocs!(ctx::Context, xs::AbstractVector{<:Processor}) = lock(ctx) do 
     append!(ctx.procs, xs)
 end
+"""
+    rmprocs!(ctx::Context, xs)
+
+Remove the specified workers `xs` from `ctx`.
+
+Workers will typically finish all their assigned tasks if scheduling is ongoing but will not be assigned new tasks after removal.
+
+Workers can be either `Processors` or the underlying process ids as `Integer`s.
+"""
+rmprocs!(ctx::Context, xs::AbstractVector{<:Integer}) = rmprocs!(ctx, map(OSProc, xs))
+rmprocs!(ctx::Context, xs::AbstractVector{<:Processor}) = lock(ctx) do 
+    filter!(p -> p ∉ xs, ctx.procs)
+end
+
--- a/src/processor.jl
+++ b/src/processor.jl
@@ -238,6 +238,7 @@ mutable struct Context
     log_sink::Any
     profile::Bool
     options
+    proc_lock::ReentrantLock
 end
 
 """
@@ -256,6 +257,7 @@ as a `Vector{Int}`.
 function Context(xs)
     Context(xs, NoOpLog(), false, nothing) # By default don't log events
 end
+Context(xs, log_sink, profile, options) = Context(xs, log_sink, profile, options, ReentrantLock())
 Context(xs::Vector{Int}) = Context(map(OSProc, xs))
 function Context()
     procs = [OSProc(w) for w in workers()]
@@ -270,4 +272,25 @@ Write a log event
 """
 function write_event(ctx::Context, event::Event)
     write_event(ctx.log_sink, event)
+end
+
+"""
+    lock(f, ctx::Context)
+
+Acquire `ctx.proc_lock`, execute `f` with the lock held, and release the lock when `f` returns.
+"""
+Base.lock(f, ctx::Context) = lock(f, ctx.proc_lock)
+
+"""
+    addprocs!(ctx::Context, xs)
+
+Add new workers `xs` to an existing `Context ctx`. 
+
+Workers will typically be assigned new tasks in the next scheduling iteration.
+
+Workers can be either `Processors` or the underlying process ids as `Integer`s.
+"""
+addprocs!(ctx::Context, xs::AbstractVector{<:Integer}) = addprocs!(ctx, map(OSProc, xs))
+addprocs!(ctx::Context, xs::AbstractVector{<:Processor}) = lock(ctx) do 
+    append!(ctx.procs, xs)
 end
--- a/src/scheduler.jl
+++ b/src/scheduler.jl
@@ -3,7 +3,7 @@ module Sch
 using Distributed
 import MemPool: DRef
 
-import ..Dagger: Context, Processor, Thunk, Chunk, OSProc, order, free!, dependents, noffspring, istask, inputs, affinity, tochunk, @dbg, @logmsg, timespan_start, timespan_end, unrelease, procs, move, choose_processor, execute!, rmprocs!
+import ..Dagger: Context, Processor, Thunk, Chunk, OSProc, order, free!, dependents, noffspring, istask, inputs, affinity, tochunk, @dbg, @logmsg, timespan_start, timespan_end, unrelease, procs, move, choose_processor, execute!, rmprocs!, addprocs!
 
 include("fault-handler.jl")
 
@@ -112,7 +112,7 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
         end
 
         # Note: worker_state may be different things for different contexts. Don't touch it out here!
-        worker_state = assign_new_workers!(ctx ,ps, state, chan, node_order, worker_state)
+        worker_state = assign_new_workers!(ctx, ps, state, chan, node_order, worker_state)
 
         if isempty(state.running)
             # the block above fired only meta tasks
--- a/src/scheduler.jl
+++ b/src/scheduler.jl
@@ -95,13 +95,7 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
     node_order = x -> -get(ord, x, 0)
     state = start_state(deps, node_order)
     # start off some tasks
-    for p in ps
-        isempty(state.ready) && break
-        task = pop_with_affinity!(ctx, state.ready, p, false)
-        if task !== nothing
-            fire_task!(ctx, task, p, state, chan, node_order)
-        end
-    end
+    worker_state = assign_new_workers!(ctx, procs, state, chan, node_order)
     @dbg timespan_end(ctx, :scheduler_init, 0, master)
 
     # Loop while we still have thunks to execute
@@ -116,6 +110,9 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
                 end
             end
         end
+
+        # Note: worker_state may be different things for different contexts. Don't touch it out here!
+        worker_state = assign_new_workers!(ctx ,ps, state, chan, node_order, worker_state)
 
         if isempty(state.running)
             # the block above fired only meta tasks
@@ -152,6 +149,20 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
         @dbg timespan_end(ctx, :scheduler, thunk_id, master)
     end
     state.cache[d]
+end
+
+function assign_new_workers!(ctx, ps, state, chan, node_order, nadded=0)
+    ps != procs(ctx) && return nadded
+    lock(ctx) do
+        for p in ps[1+nadded:end]
+            isempty(state.ready) && break
+            task = pop_with_affinity!(ctx, state.ready, p, false)
+            if task !== nothing
+                fire_task!(ctx, task, p, state, chan, node_order)
+            end
+        end
+        return length(ps)
+    end
 end
 
 function pop_with_affinity!(ctx, tasks, proc, immediate_next)
--- a/src/scheduler.jl
+++ b/src/scheduler.jl
@@ -3,7 +3,7 @@ module Sch
 using Distributed
 import MemPool: DRef
 
-import ..Dagger: Context, Processor, Thunk, Chunk, OSProc, order, free!, dependents, noffspring, istask, inputs, affinity, tochunk, @dbg, @logmsg, timespan_start, timespan_end, unrelease, procs, move, choose_processor, execute!
+import ..Dagger: Context, Processor, Thunk, Chunk, OSProc, order, free!, dependents, noffspring, istask, inputs, affinity, tochunk, @dbg, @logmsg, timespan_start, timespan_end, unrelease, procs, move, choose_processor, execute!, rmprocs!
 
 include("fault-handler.jl")
 
@@ -125,7 +125,7 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
                 @warn "Worker $(proc.pid) died on thunk $thunk_id, rescheduling work"
 
                 # Remove dead worker from procs list
-                filter!(p->p.pid!=proc.pid, ctx.procs)
+                rmprocs!(ctx, [proc])
                 ps = procs(ctx)
 
                 handle_fault(ctx, state, state.thunk_dict[thunk_id], proc, chan, node_order)
@@ -140,7 +140,7 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
 
         @dbg timespan_start(ctx, :scheduler, thunk_id, master)
         immediate_next = finish_task!(state, node, node_order)
-        if !isempty(state.ready)
+        if !isempty(state.ready) && !shall_remove_worker(ctx, proc, ps, immediate_next) 
             thunk = pop_with_affinity!(Context(ps), state.ready, proc, immediate_next)
             if thunk !== nothing
                 fire_task!(ctx, thunk, proc, state, chan, node_order)
@@ -151,17 +151,25 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
     state.cache[d]
 end
 
-function assign_new_workers!(ctx, ps, state, chan, node_order, nadded=0)
-    ps != procs(ctx) && return nadded
+function assign_new_workers!(ctx, ps, state, chan, node_order, assignedprocs=[])
+    ps !== procs(ctx) && return assignedprocs
     lock(ctx) do
-        for p in ps[1+nadded:end]
+        # Must track individual procs to handle the case when procs are removed
+        for p in setdiff(ps, assignedprocs)
             isempty(state.ready) && break
             task = pop_with_affinity!(ctx, state.ready, p, false)
             if task !== nothing
                 fire_task!(ctx, task, p, state, chan, node_order)
             end
         end
-        return length(ps)
+        return copy(ps)
+    end
+end
+
+function shall_remove_worker(ctx, proc, ps, immediate_next) 
+    ps !== procs(ctx) && return false
+    return lock(ctx) do 
+        proc ∉ procs(ctx)
     end
 end
 
--- a/test/scheduler.jl
+++ b/test/scheduler.jl
@@ -66,15 +66,15 @@ end
     end
     @everywhere (pop!(Dagger.PROCESSOR_CALLBACKS); empty!(Dagger.OSPROC_CACHE))
 
-    @testset "Add new workers" begin
-        # Test that we can add new workers to an ongoing task.
+    @testset "Modify workers in running job" begin
+        # Test that we can add/remove workers while scheduler is running.
         # As this requires asynchronity a Condition is used to stall the tasks to 
-        # ensure workers are actually added while the scheduler is working 
-        using Distributed
-  
+        # ensure workers are actually modified while the scheduler is working 
+
         setup = quote
            using Dagger, Distributed
            # Condition to guarantee that processing is not completed before we add new workers
+           # Note: c is used in expressions below
            c = Condition()
            function testfun(i)
                i < 4 && return myid()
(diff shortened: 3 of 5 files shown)


DrChainsaw:
Some attempts as benchmarking. Does not point towards any issue with overhead at all, but maybe I'm not doing it right.

<details><summary>Current master</summary><p>

```julia

julia> using Dagger, Distributed, BenchmarkTools
[ Info: Precompiling Dagger [d58978e5-989f-55fb-8d15-ea34adc7bf54]

julia> ts = delayed(vcat)((delayed(i -> i)(i) for i in 1:1000)...);

julia> @benchmark collect($ts)
BenchmarkTools.Trial:
  memory estimate:  607.76 MiB
  allocs estimate:  4029962
  --------------
  minimum time:     327.600 ms (18.88% GC)
  median time:      341.593 ms (17.17% GC)
  mean time:        342.693 ms (17.26% GC)
  maximum time:     365.156 ms (14.88% GC)
  --------------
  samples:          15
  evals/sample:     1

julia> addprocs(100, exeflags="--project"); # Machine only has 4 cores, so this is not expected to be efficient :)

julia> @everywhere using Dagger

julia> @benchmark collect($ts)
BenchmarkTools.Trial:
  memory estimate:  661.37 MiB
  allocs estimate:  5146505
  --------------
  minimum time:     6.912 s (1.05% GC)
  median time:      6.912 s (1.05% GC)
  mean time:        6.912 s (1.05% GC)
  maximum time:     6.912 s (1.05% GC)
  --------------
  samples:          1
  evals/sample:     1
```

</p></details>

<details><summary>This PR</summary><p>

```julia
julia> using Dagger, Distributed, BenchmarkTools

julia> ts = delayed(vcat)((delayed(i -> i)(i) for i in 1:1000)...);

julia> @benchmark collect($ts)
BenchmarkTools.Trial:
  memory estimate:  608.86 MiB
  allocs estimate:  4053113
  --------------
  minimum time:     281.962 ms (17.93% GC)
  median time:      305.351 ms (17.61% GC)
  mean time:        318.522 ms (17.14% GC)
  maximum time:     401.144 ms (15.41% GC)
  --------------
  samples:          16
  evals/sample:     1

julia> addprocs(100, exeflags="--project"); # Machine only has 4 cores, so this is not expected to be efficient :)

julia> @everywhere using Dagger

julia> @benchmark collect($ts)
BenchmarkTools.Trial:
  memory estimate:  691.25 MiB
  allocs estimate:  5706156
  --------------
  minimum time:     6.118 s (1.11% GC)
  median time:      6.118 s (1.11% GC)
  mean time:        6.118 s (1.11% GC)
  maximum time:     6.118 s (1.11% GC)
  --------------
  samples:          1
  evals/sample:     1
```

</p></details>


<details><summary>When does overhead from setdiff become significant?</summary><p>

```julia
julia> function test_setdiff(sizes)
       for s in sizes
       @info "Test size $s"
       arr = collect(1:s)
       @btime setdiff($arr, $arr)
       end
       end
test_setdiff (generic function with 1 method)

julia> test_setdiff(10 .^ (0:6))
[ Info: Test size 1
  309.619 ns (7 allocations: 608 bytes)
[ Info: Test size 10
  470.769 ns (7 allocations: 672 bytes)
[ Info: Test size 100
  3.475 μs (13 allocations: 7.40 KiB)
[ Info: Test size 1000
  38.500 μs (14 allocations: 53.91 KiB)
[ Info: Test size 10000
  564.299 μs (12 allocations: 222.90 KiB)
[ Info: Test size 100000
  8.573 ms (17 allocations: 4.14 MiB)
[ Info: Test size 1000000
  165.722 ms (17 allocations: 34.63 MiB)
```
</p></details>

jpsamaroo:
>The option to provide a single worker through options makes it awkward to use procs(ctx) in the scheduling loop so for now ps is still being used. The option of single worker is also handled through a number of ps === procs(ctx) checks which is maybe not so nice. Why would anyone use it over creating a context with just one proc in it?

Yeah, I never really thought through `single` when I included it. If you want to remove it, I'd be fine with that. Just make sure that the docstrings get updated.

> The faulthandler is currently a bit scary as it just plucks a random worker. Shouldn’t it pick one from the context? What if the picked worker is not setup with everywhere yet? This took me longer than I’d like to admit to figure out as the shall_remove_proc function did not play well with this at all: The whole thing just stalled when running the fault-tolerance tests since new workers where not in the context so no tasks were scheduled for them. :)

The only good thing about the fault handler is that it sometimes works, not that it's a good implementation :smile: I wrote it before I fully understood how Dagger's `Context` worked, so it does some things wrong. Please don't hesitate to fix it!

> I can add a section to the docs as well as soon as the implementation has converged.

Please do, that would be very helpful.

> Shall I make addprocs! part of the public API?

Yes

> Should it be called addprocs to align with Distributed?

No, since it's a mutating function. If it returned a new `Context` with those procs added then I'd say yes (but that's not the point of this PR).

> Same as above for rmprocs!.

Same answers as for addprocs!

> I don’t love how complicated the testcases for this are. Any suggestions on how to simplify them?

Not currently, but I'll think on it.

src/processor.jl
```diff
@@ -271,3 +273,38 @@ Write a log event
 function write_event(ctx::Context, event::Event)
     write_event(ctx.log_sink, event)
 end
+
+"""
+    lock(f, ctx::Context)
+
+Acquire `ctx.proc_lock`, execute `f` with the lock held, and release the lock when `f` returns.
+"""
+Base.lock(f, ctx::Context) = lock(f, ctx.proc_lock)
```
jpsamaroo, line 282: This also needs to be applied to `procs(ctx)` for correctness:

```julia
procs(ctx) = lock(ctx) do
    copy(ctx.procs)
end
```

test/scheduler.jl
```diff
@@ -66,31 +66,96 @@ end
...
-        job = @async collect(Context(ps1), ts);
-    
-        ps2 = addprocs(2, exeflags="--project");
-    
-        while !istaskdone(job)
-            @everywhere ps1 notify(c)
+
+        @testset "Add new workers" begin
+        ps = []
+        try
```
jpsamaroo, line 88: Indentation

test/scheduler.jl
```diff
@@ -66,31 +66,96 @@ end
...
+
+        @testset "Add new workers" begin
+        ps = []
+        try     
+                ps1 = addprocs(2, exeflags="--project");
+                push!(ps, ps1)
+
+                @everywhere vcat(ps1, myid()) $setup
+        
+                ts = delayed(vcat)((delayed(testfun)(i) for i in 1:10)...);
```
jpsamaroo, line 94: Semicolons here and below are unnecessary

test/scheduler.jl
```diff
@@ -66,31 +66,96 @@ end
...
-        ps2 = addprocs(2, exeflags="--project");
-    
-        while !istaskdone(job)
-            @everywhere ps1 notify(c)
+
+        @testset "Add new workers" begin
+        ps = []
+        try     
+                ps1 = addprocs(2, exeflags="--project");
+                push!(ps, ps1)
```
jpsamaroo, line 90: I think you mean `append!`, otherwise you have a `Vector{Vector{Int}}`. Apparently passing a vector of the wrong type to `Distributed.rmprocs` silently fails, which is a terrible thing...

test/scheduler.jl
```diff
@@ -66,31 +66,96 @@ end
...
+                ps2 = addprocs(2, exeflags="--project");
+                push!(ps, ps2)
+
+                ps3 = addprocs(2, exeflags="--project")
+                push!(ps, ps3)
+                @everywhere ps3 $setup
+                Dagger.addprocs!(ctx, ps3)
+                @test length(procs(ctx)) == 4
+        
+                while !istaskdone(job)
```
jpsamaroo, line 113: A single `notify` call is sufficient, and then you just rely on the below `fetch(job)` to wait on it to complete.

test/scheduler.jl
```diff
@@ -66,31 +66,96 @@ end
...
+                    @everywhere ps1 notify(c)
+                    @everywhere ps3 notify(c)
+                end
+                @test fetch(job) isa Vector
+                @test fetch(job) |> unique |> sort == vcat(ps1, ps3)
```
jpsamaroo, line 118: Technically this test will start failing on future PRs when we start scheduling multiple thunks onto a single worker. However we don't have a decent way to ensure that `ps3` will be scheduled on right now (`options.single` only works for workers in our context), so this is fine for now.

test/scheduler.jl
```diff
@@ -66,31 +66,96 @@ end
...
+
+            finally
+                wait(rmprocs(ps))
+            end
         end
-        @test fetch(job) |> unique |> sort == ps1
 
-        wait(rmprocs(vcat(ps1,ps2)))
+        @testset "Remove workers" begin
+        ps = []
```
jpsamaroo, line 126: Same reviews for "Adding workers" applies to this testset.

jpsamaroo (changes requested):
Looking good, just a few nitpicks.

jpsamaroo:
Actually, I think we need a better way to workaround `single`, other than removing it. Let me think on it for a bit.

src/processor.jl
DrChainsaw: Indeed it does. Requires a bit of a rewrite of `compute_dag` but nothing too serious.

test/scheduler.jl
DrChainsaw: Will fix!

test/scheduler.jl
DrChainsaw: Will fix!

test/scheduler.jl
DrChainsaw: yikes!

test/scheduler.jl
DrChainsaw: I think a single call will stall when the function is called for the next index. Lemme see if I can just rewrite with a busy wait and a boolean instead. Or perhaps a lock...

test/scheduler.jl
DrChainsaw: I'll add a comment about that a least. It is a bit of a integration type test which does indeed make it a bit fragile.

test/scheduler.jl
DrChainsaw: Will fix!

DrChainsaw:
> Actually, I think we need a better way to workaround `single`, other than removing it. Let me think on it for a bit.

Ok, I could just add the options as input to `assign_new_workers!` and check them there again.

test/scheduler.jl
jpsamaroo: Oh sorry, you're completely right about that! No need to make a change then.

DrChainsaw:
I think all comments except what to do with the fault handler is addressed by the above commits. Using a defensive copy opened up for just checking the options whenever one needs to loop over workers in `collect_dag` so I think there is nothing more needed there unless you think of something better.

I added a parallel task to poll for new workers as you would otherwise need to wait for a task to finish before checking. I left the check in the loop as well as the new testcases failed otherwise and I didn't want to add arbitrary waiting times to them.

> The only good thing about the fault handler is that it sometimes works, not that it's a good implementation 😄

Well, working code is often better than no code :)

Checking against context or options in the fault handler should be straight forward, but what should one do if there are no more workers to try? Current code will probably end up waiting for `take!(chan)` indefinitely so maybe there should be an error when nothing is running and there are no workers. Technically the user could decide to add some more, but I don't think it is good to assume they will. If the check is dispatched on context or (perhaps better) options one can allow users to specify that this is wanted behaviour through custom structs.

W.r.t fault handling, maybe just trying a worker at random is a worthwhile last resort though. If options or context is involved then advanced users can maybe do things like starting a processes (and `@everywhere` them) if they want to.

DrChainsaw:
Not sure if the async launching of new workers is too intrusive. Another option is to add a short sleep and a continue if the channel is empty instead of `take!`, but that will add unavoidable latency in the scheduling loop.

A shower thought about the fault handling: Wouldn't it make sense if the behaviour when there are no more workers is to return the current partially computed state. The user can then just restart (after restarting the workers) and it also gives at least a theoretical chance of fixing any issues with the computation without throwing away results. Do you think this is feasible to do?

I guess that if the intermediate result is so too big to fit in the host it might be killed right away so without explicitly checkpointing to files there is perhaps no way to be safe about it.

jpsamaroo:
> Checking against context or options in the fault handler should be straight forward, but what should one do if there are no more workers to try?

An error is the only answer here, especially since that probably means that the user's code itself is causing the workers to terminate abnormally.

> W.r.t fault handling, maybe just trying a worker at random is a worthwhile last resort though.

As long as that worker is in Context, yes. It was an oversight on my part to not use workers from the context in the fault handler.

src/scheduler.jl
```diff
@@ -125,8 +119,8 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
                 @warn "Worker $(proc.pid) died on thunk $thunk_id, rescheduling work"
 
                 # Remove dead worker from procs list
+                # Not sure what is desired behaviour if option.singleworker is set...
```
jpsamaroo, line 136: We should error out at this point if `single` is set.

src/scheduler.jl
```diff
@@ -89,9 +89,20 @@ function compute_dag(ctx, d::Thunk; options=SchedulerOptions())
     node_order = x -> -get(ord, x, 0)
     state = start_state(deps, node_order)
     # start off some tasks
+    # Note: worker_state may be different things for different contexts. Don't touch it out here!
     procs_state = assign_new_procs!(ctx, state, chan, node_order)
     @dbg timespan_end(ctx, :scheduler_init, 0, master)
 
+    # Check periodically for new workers in a parallel task so that we don't accidentally end up
```
jpsamaroo, line 96: I don't think is necessary or desired; if there's the possibility that we could be waiting on `take!(chan)` for completed work that doesn't exist, then we should do a check that `!isempty(state.running)` prior to waiting. That would ensure at least one thunk is in-flight and will eventually return to us.

Also, currently this will run forever (even after the scheduler completes the DAG).

docs/src/processors.md
```diff
@@ -70,3 +70,48 @@ complicated detection and recovery process, including multiple master
...
+# Dummy task to wait for 0.5 seconds and then return the id of the worker
+ts = delayed(vcat)((delayed(i -> (sleep(0.5); myid()))(i) for i in 1:20)...);
+
+ctx = Context();
+# Scheduler is blocking, so we need a new task to add workers while it runs
+job = @async collect(ctx, ts); 
+
+# Lets fire up some new workers
+ps2 = addprocs(2, exeflags="--project");
+@everywhere using Distributed, Dagger
```
jpsamaroo, line 108:
```suggestion
@everywhere ps2 using Distributed, Dagger
```

jpsamaroo (changes requested):

jpsamaroo:
Also, we need to address the second part of my comment, about ensuring we don't have anything cached on removed workers:

https://github.com/JuliaParallel/Dagger.jl/issues/145#issuecomment-702364557

src/scheduler.jl
DrChainsaw: Hmmm, it was intended as an optimization more than handling when chan is never populated.

In my use case which triggered the desire for this functionality, I have about 400 files which each take 2-30 minutes to process. Assume that the cluster scheduler starts by dispatching say 10 out of 400 jobs. If I'm really unlucky those 10 workers each get to work on a 30 minute file and even if I get more workers in the meantime no new tasks will be started by the scheduler until one of those 10 workers are done with their current task since `take!` blocks the thread. Note that it might take way more than 30 minutes to get all 400 jobs up and running, maybe several days if the cluster is heaviliy loaded.

I don't love that the loop is littered with locks now so I can try something else if that is the problem, but I think the use case is there. Or have I misunderstood something?

src/scheduler.jl
DrChainsaw: Ok, will fix!

DrChainsaw:
> Also, we need to address the second part of my comment, about ensuring we don't have anything cached on removed workers:
> 
> [#145 (comment)](https://github.com/JuliaParallel/Dagger.jl/issues/145#issuecomment-702364557)

Even if workers are not removed from Distributed tracking? Afaik, they are still referenced by their Chunks and fetching from them works as normal. This was all part of the "no promises made" for `rmprocs!`.

src/scheduler.jl
DrChainsaw: I put the same condition in the asynch task as in the main scheduling loop, so it should stop when there is nothing to compute and nothing running. Will double check that now just in case...

DrChainsaw:
Added the following w.r.t fault handling:
Fault handling is not attempted if `options.single > 0`. Instead an error is thrown.
Fault handling in other cases will use `procs(ctx)` and throw if no more processes are found

Scheduler will not throw if no processes are found as it can be assumed this is a deliberate move from the user.

src/scheduler.jl
jpsamaroo: > Also, currently this will run forever (even after the scheduler completes the DAG).

> I put the same condition in the asynch task as in the main scheduling loop, so it should stop when there is nothing to compute and nothing running. Will double check that now just in case...

Sorry, I misread the implementation, you're right!

> In my use case which triggered the desire for this functionality, I have about 400 files which each take 2-30 minutes to process. Assume that the cluster scheduler starts by dispatching say 10 out of 400 jobs. If I'm really unlucky those 10 workers each get to work on a 30 minute file and even if I get more workers in the meantime no new tasks will be started by the scheduler until one of those 10 workers are done with their current task since take! blocks the thread. Note that it might take way more than 30 minutes to get all 400 jobs up and running, maybe several days if the cluster is heaviliy loaded.

This is an interesting case that I had considered but never fixed in the scheduler. I think the long-term solution is to split off the receiving of completed work from the channel into an `@async` or `@spawn` task, and having one or more separate tasks working on sending work to the workers. This seems to accomplish that, so I'm on board with your implementation.

jpsamaroo:
> Even if workers are not removed from Distributed tracking? Afaik, they are still referenced by their Chunks and fetching from them works as normal. This was all part of the "no promises made" for rmprocs!.

Oh I see, that's fine. Maybe in the future we can add a mechanism to ensure that the worker is completely unused by the scheduler so that we do a `Distributed.rmprocs` on it.

> Ok, this should be pretty straight forward to do. Shall the error be thrown only at failure or at any time the scheduler finds it has more work to do but no workers?

The former should be fine, since if `procs(ctx)` returns an empty vector during regular scheduling, the code using its results won't be able to select a proc for scheduling (so we'd get an error regardless).

src/scheduler.jl
DrChainsaw: Awesome! I will take another pass to see if I can reduce the amount of moving parts. Afaik, I need to lock down all cases of pop_with_affinity! to prevent that the same task is scheduled twice, right? I see now that I have missed doing this for the fault handling.

I'm thinking that the only place where it is necessary to check for new workers is when waiting for `take!(chan)`, so perhaps either keep the lock locked everywhere in the loop except there or spawn the task right before waiting if the channel is empty and then stop it (with a flag) right after something popped out of the channel. 

Thoughts on this?

DrChainsaw:
> Oh I see, that's fine. Maybe in the future we can add a mechanism to ensure that the worker is completely unused by the scheduler so that we do a Distributed.rmprocs on it.

Sounds good. This PR is big enough as it is and I'm in no need for this. I'll see if I can add a hint about this in the docs.

> The former should be fine, since if procs(ctx) returns an empty vector during regular scheduling, the code using its results won't be able to select a proc for scheduling (so we'd get an error regardless).

Not being able to select a proc in scheduling results in waiting for `take!(chan)` and the add workers thread idenfinitely. I'm thinking this is kinda ok as a user need to put some effort in to end up here so chances are they know what they are doing.

jpsamaroo:
> Not being able to select a proc in scheduling results in waiting for `take!(chan)` and the add workers thread idenfinitely. I'm thinking this is kinda ok as a user need to put some effort in to end up here so chances are they know what they are doing.

I prefer errors over hangs any day; could we do an `isempty` check in the scheduler to make sure we don't ever encounter a hang?

DrChainsaw:
> I prefer errors over hangs any day; could we do an isempty check in the scheduler to make sure we don't ever encounter a hang?

I agree in general and I will add it. Again, the user has to kinda go out if their way to trigger it which is why I thought it could be worth an exception from the general guideline of faling fast. I'll add a hook for it so that it is possible to circumvent the error if really needed.

I'll submit what is hopefully the last batch of changes later tonight.

DrChainsaw:
I think thats the last of them. I will do some tests in my environment and then post here if I think I'm done (barring further review comments ofc).

I reworked the mechanism for launching new tasks so that it is started right before `take!(chan)` and stopped right after. This feels safer than forcing everyone who will ever contribute to the scheduler to keep the background task in mind when hacking about.

The scheduling loop now asserts that the context still has procs (unless options.single is used).

Btw, I tried to add some testing for the empty procs case in the fault handler, but I had a hard time to get the fault handling to work outside the existing cases for the `kill_lazy` case. Most of my attempts resulted in the scheduling loop exiting without error and then throwing an error because the result was not in the cache.

DrChainsaw:
Ok, use case checks out. I'm all set if you are!

jpsamaroo (approved):

DrChainsaw:
Thanks alot for reviewing (which I deem to be even harder work :) ) and for pulling!

PR #151, opened 2020-10-16
Make benchmarking script simpler

jpsamaroo:
The old benchmarking script assumed that the file I/O capabilities of Dagger still worked well, which is not the case. It also scaled the size of the benchmark based on the number of workers, and the `judge` call didn't work because it didn't specify the metric.

Eventually, I'd like to run this automatically as a CI job, but until then I'll use it manually to test scheduler changes.


jpsamaroo:
Superseded by #169

PR #158, opened 2020-10-29
Add TLS accessor for processor in Thunk

jpsamaroo:
Helpful for dispatching on Processor from within a Thunk

### DIFF
--- a/src/processor.jl
+++ b/src/processor.jl
@@ -318,4 +318,4 @@ rmprocs!(ctx::Context, xs::AbstractVector{<:Processor}) = lock(ctx) do
 end
 
 "Gets the current processor executing the current thunk."
-thunk_processor() = task_local_storage(:processor)
+thunk_processor() = task_local_storage(:processor)::Processor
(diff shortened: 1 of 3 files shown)


src/processor.jl
```diff
@@ -310,3 +316,6 @@ rmprocs!(ctx::Context, xs::AbstractVector{<:Integer}) = rmprocs!(ctx, map(OSProc
 rmprocs!(ctx::Context, xs::AbstractVector{<:Processor}) = lock(ctx) do
     filter!(p -> p ∉ xs, ctx.procs)
 end
+
+"Gets the current processor executing the current thunk."
+thunk_processor() = task_local_storage(:processor)
```
vchuravy, line 321: Maybe add a type check to make this type stable? Or is it possible to be many different kind of Processors? Even checking that it is a subtype of an Abstact processor might be helpful.

jpsamaroo: Yeah we can only really assert `::Processor`. I just pushed the typeassert.

Issue #159, opened 2020-10-29
UI: Add Gantt chart


jpsamaroo:
Ideally also showing the breakdown of the processor hierarchy

jpsamaroo:
Implemented in #169 and improved in #278

PR #169, opened 2020-11-30
Add Gantt chart using Mux and Luxor

jpsamaroo:
Todo:

- [ ] Add `scheduler_init` vertical red line
- [x] Get profiling working well
- [x] Require FFMPEG to allow offline video generation
- [ ] "Bunch up" multiple proc lines
- [x] Docs
- [ ] Try to test this


jpsamaroo:
Closes #159

jpsamaroo:
I've merged in the benchmark changes here because as I developed them they became very intertwined, such that the benchmarking script now also generates the Gantt chart and Profile flamegraph offline by default and stores them with the results .jld for analysis.

vchuravy:
Looking at https://mathdugre.me/paper-big-data-engines/dask-delayed-inc-baseline.html as an example they use different colours to differentiate between "compute", "read", "write", but keep the the colours between tasks constant.

For us I think it would make sense to have "schedule", "communication", "computation", "checkpointing", as categories.
The example above comes from dask and they are using Bokeh, which allows them to attach meta information and display that on hover.

jpsamaroo:
I'm going to merge this without tests for right now, because this entire infrastructure is pretty experimental, and doesn't yet cover all the cases that the original benchmarking script used to cover (although this one actually works). I'd especially appreciate any artistic input (w.r.t Gantt chart, flamechart, and benchmark plots visualization) as issues/PRs going forward!

jpsamaroo:
Thanks @vchuravy for all the help with directing this work!

PR #171, opened 2020-12-01
An ocean of thunks

jpsamaroo:
Add processor utilization tracking
Add support for selecting processors based on a user-specified function

Todo: Allow the scheduler to split the DAG and send them to workers to process


jpsamaroo:
Superseded by #202

PR #176, opened 2020-12-10
Fix readme badges, print out version in nightly

christopher-dG:
There's no coverage submission from within Buildkite so I took away that badge. I can try to get it working later though.


jpsamaroo:
Thanks!

Issue #84, opened 2018-03-20
Scheduler not dispatching to workers


tk3369:
This may be the same issue as #71 .

I have 3 workers and  I'm using Julia 0.6.2.

This simple graph has 3 leaf thunks (`x2`) and 1 master thunk (`x3`).  It works properly.
```
julia> x2 = [delayed(_ -> (sleep(2); println(i)))(nothing) for i in 1:3]
3-element Array{Dagger.Thunk,1}:
 *29*
 *30*
 *31*

julia> x3 = delayed((x...) -> println("done: ", x))(x2...)
*32*

julia> @time compute(x3)
	From worker 3:	2
	From worker 4:	3
	From worker 2:	1
	From worker 2:	done: (nothing, nothing, nothing)
  2.320594 seconds (21.23 k allocations: 1.159 MiB)
Dagger.Chunk{Void,MemPool.DRef}(Void, Dagger.UnitDomain(), MemPool.DRef(2, 39, 0x0000000000000000), false)
```

This case is broken.  The only difference is that the new leaf thunk (`x1`) was added for which `x2` thunks depend on.

```
julia> x1 = delayed(x -> (sleep(2); println("root")))(1)
*33*

julia> x2 = [delayed(_ -> (sleep(2); println(i)))(x1) for i in 1:3]
3-element Array{Dagger.Thunk,1}:
 *34*
 *35*
 *36*

julia> x3 = delayed((x...) -> println("done: ", x))(x2...)
*37*

julia> @time compute(x3)
	From worker 2:	root
	From worker 2:	1
	From worker 2:	2
	From worker 2:	3
	From worker 2:	done: (nothing, nothing, nothing)
  8.153534 seconds (31.29 k allocations: 1.678 MiB)
Dagger.Chunk{Void,MemPool.DRef}(Void, Dagger.UnitDomain(), MemPool.DRef(2, 44, 0x0000000000000000), false)
```

jpsamaroo:
Based on adding some debug statements to the scheduler, I think what's happening here is the following:

- `x1` runs to completion.
- `x2[1]` begins executing and is placed in `state.running`, and `x2[2]` and `x2[3]` are placed in `state.ready`. This is the first misstep by the scheduler, as it's clear that `x2[2]` and `x2[3]` can be run. Probably the same issue as #53 
[...]

PR #177, opened 2020-12-10
Remove immediate_next fast-path

jpsamaroo:
When we "finish" a thunk (mark it completed), we immediately assign the processor some work based on how close the data is to the worker. Sometimes, we'll even refuse to schedule if we don't think the worker should have a given piece of work. Over time, the worker accrues so much of the work, that it also accrues most of the output data, and then ends up getting stuck being the only worker doing anything. In a sense, the scheduler is being overly conservative about data transfer costs, and failing to consider compute costs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .*~
 .*sw?
 Manifest.toml
+*.jld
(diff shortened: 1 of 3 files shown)


jpsamaroo:
*Finally* fixes #84:
```julia
julia> @time compute(x3)
root
2
1
3
done: (nothing, nothing, nothing)
  4.006687 seconds (1.52 k allocations: 99.953 KiB)
Dagger.Chunk{Any, MemPool.DRef, Dagger.ThreadProc}(Nothing, UnitDomain(), MemPool.DRef(1, 9, 0x0000000000000000), Dagger.ThreadProc(1, 2), false)
```

PR #178, opened 2020-12-15
Actually test on nightly

christopher-dG:
The previous version should work, but apparently it doesn't, and the nightly test is running Julia 1.5 (you can see it in the stacktrace [here](https://buildkite.com/julialang/dagger-dot-jl/builds/22#d678a78e-8343-42c8-b820-28497f4652f3/18-248)).

This one ought to do the trick.

edit: this revealed that the download of nightly is broken, don't merge this until I fix that

edit2: working now


jpsamaroo:
Thanks!

PR #210, opened 2021-04-08
Thunk cost estimation, chunk caching, benchmark updates

### LINKED ISSUE #205: Auto-detect thunk utilization
For array operations in particular, we should we able to guess how many threads a given operation will use, and so we should be able to populate `procutil` automatically in some cases. This should be programmable with multiple dispatch, of course.

jpsamaroo:
This PR adds runtime estimation of thunk costs (per signature) to get as close to max utilization as possible (without impacting total runtime). It also adds chunk data caches to workers, which will be freed by the scheduler once they're no longer needed.

~Depends on https://github.com/JuliaData/MemPool.jl/pull/49~ Not doing this for now, wrong API for performance reasons.

Closes #205

Todo:
- [x] Parameterize function cost cache on argument types
- [x] Cache Chunk arguments per-process
- [x] Also key chunk cache on processor
- [x] Track Chunk usage and evict from caches ASAP
- [x] Test caching behavior
- [ ] ~Scale thunk cost by number of running thunks on same processor~ Something for later, possibly
- [ ] ~Add lots more benchmarks from https://blog.dask.org/2017/07/03/scaling~ https://github.com/JuliaParallel/Dagger.jl/issues/220
- [x] Confirm we benchmark better than master

### DIFF
--- a/src/sch/Sch.jl
+++ b/src/sch/Sch.jl
@@ -438,7 +438,7 @@ function Base.show(io::IO, se::SchedulingException)
     print(io, "SchedulingException ($(se.reason))")
 end
 
-const CHUNK_CACHE = Dict{Chunk,Any}()
+const CHUNK_CACHE = Dict{Chunk,Dict{Processor,Any}}()
 
 function schedule!(ctx, state, procs=procs_to_use(ctx))
     lock(state.lock) do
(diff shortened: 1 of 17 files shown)


src/sch/Sch.jl
```diff
@@ -146,30 +231,34 @@ function isrestricted(task::Thunk, proc::OSProc)
...
             state.worker_capacity[p.pid][T] = capacity(proc, T)
         end
         state.worker_pressure[p.pid][OSProc] = 0
         state.worker_capacity[p.pid][OSProc] = 0
     end
-    cap = remotecall_fetch(capacity, p.pid)
+    cap = remotecall(capacity, p.pid)
     @async lock(state.lock) do
-        state.worker_capacity[p.pid] = cap
+        state.worker_capacity[p.pid] = fetch(cap)
```
vchuravy, line 259: Can we do the fetch before we acquire the lock?

PR #217, opened 2021-05-20
Add Dagger.spawn, allow on any node

### LINKED ISSUE #215: Allow `Dagger.@spawn` on non-head node
We can just do a `remotecall` in the macro if we aren't on the head node.

jpsamaroo:
@tkf @DhairyaLGandhi

Closes #215

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Dagger"
 uuid = "d58978e5-989f-55fb-8d15-ea34adc7bf54"
-version = "0.11.2"
+version = "0.11.3"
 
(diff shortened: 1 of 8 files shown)


jpsamaroo:
Slight complication: if we want to do nested `@spawn`, we need some way to ensure that the thunk we're trying to `wait`/`fetch` gets scheduled before our thunk completes. Logically, the right thing to do is to teach the scheduler to temporarily reduce the spawning thunk's pressure to 0 so that the new thunk has capacity to execute (and once we finish waiting, we set it back to its original value).

PR #221, opened 2021-06-03
Optimize thunk placement for input affinity

jpsamaroo:
The previous algorithm only accounted for how many thunks could fit on a single processor, without regard for where that thunk's inputs are coming from (thus causing lots of potentially unnecessary network transfers). The new algorithm is simply placed before the existing one, and tries to find the processor which contains the most amount of input data for a given thunk. If all processors containing that thunk's input data are at capacity, we simply fall back to the old algorithm.

@shashi this should make you happy :slightly_smiling_face: 

Todo:
- [ ] Amortize some costs when calculating chunk affinity and rejecting processors
- [ ] Benchmark better than master

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,7 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 Colors = "0.10, 0.11, 0.12"
-MemPool = "0.3.4"
+MemPool = "0.3.5"
 Requires = "1"
 StatsBase = "0.28, 0.29, 0.30, 0.31, 0.32, 0.33"
 julia = "1.0"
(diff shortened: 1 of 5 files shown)


vchuravy:
What is the increase in scheduling overhead?

jpsamaroo:
I haven't yet gotten any meaningful benchmark runs, but under low load and a small cluster size it shouldn't be significant. I need to think about ways to improve this approach so that we can minimize the cost per thunk a bit more, since this approach isn't very efficient.

jpsamaroo:
This PR now fixes memory leaks introduced by the dynamic scheduler control PR. These leaks occurred because the dynamic listener tasks in the scheduler captured the very large `ComputeState` object, and were not successfully terminated with the scheduler. If you previously ran into OOM issues with Dagger, please try running your code with this PR.

PR #222, opened 2021-06-25
Fix dynamic_fib

tkf:
Parallel dynamic fib requires calling `dynamic_fib(n-2)` before `fetch`. Otherwise, all recursions are done serially.

Maybe it was me who sent this code to @jpsamaroo :)

### DIFF
--- a/test/thunk.jl
+++ b/test/thunk.jl
@@ -5,7 +5,8 @@ import Dagger: @par, @spawn, spawn
     function dynamic_fib(n)
         n <= 1 && return n
         t = Dagger.spawn(dynamic_fib, n-1)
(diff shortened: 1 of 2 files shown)


test/thunk.jl
```diff
@@ -147,6 +148,6 @@ end
         @test fetch(a) == 3
 
         # Mild stress-test
-        @test fetch(dynamic_fib(10)) == 55
```
tkf, line 150: `fetch` here was redundant (it was using the method `fetch(x) = x`)

jpsamaroo:
I should probably be able to fix the hangs with #223, which fixes some races in eager thunk submission and scheduling.

jpsamaroo:
Thanks!

PR #223, opened 2021-06-28
Fix memory leaks in eager API

jpsamaroo:
We need to ensure that we allow eager-allocated `Thunk`s to become unreferenced by the scheduler when they're no longer needed, so that they and their inputs can be subject to GC.

This also fixes a memory leak caused by the dynamic listener not terminating with its associated scheduler, resulting in a leak of the entire `ComputeState`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Dagger"
 uuid = "d58978e5-989f-55fb-8d15-ea34adc7bf54"
-version = "0.11.5"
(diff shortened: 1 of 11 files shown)


jpsamaroo:
Fixes #225

jpsamaroo:
I've merged this to fix a number of bugs, but it's not yet at 100% because we still get some memory leaks (hence why we have `@test_broken` statements).

Issue #214, opened 2021-05-14
Revitalize file I/O, aka swap-to-disk


jpsamaroo:
It's not actively tested or maintained, so we have to decide if we want to remove it, or if we want to modernize it and start testing that it works for out-of-core.

shashi:
We should remove it and use FileTrees.jl

jpsamaroo:
Could FileTrees gain some logic to allow saving/loading a filetree from/to a DArray? That would mimic some of the code we're removing.

shashi:
Ehh maybe that's one level too extra of an abstraction.

You can couple of functions behind `@requires FileTrees` here which just load and save arrays by turning them into FileTrees.

shashi:
Ideally this would move to DaggerArrays and then you wouldn't need the `@requires`

jpsamaroo:
Alternatively, I think everyone could benefit from the scheduler having a basic understanding of memory pressure and have the option for data to be moved to disk when we're nearing an OOM situation (whether by mmap or Ser/Des). This could tie in nicely with the checkpointing API: even if checkpointing isn't enabled by the user, the scheduler could still decide to checkpoint (and simultaneously evict data from memory) if the data is bulky and it's expected that it won't be accessed in the short term.

jpsamaroo:
Here's my plans for a "storage subsystem" that I'm working on:
- `Storage` abstract type, currently a subtype of `Processor` to let it fit the `move()` dispatch
- Memory/disk capacity awareness (per `Storage` object), communicated to (and estimated by) the Sch scheduler
- Memory usage tracking per-signature during execution via `Base.gc_num` (although this is inaccurate, being for all threads)
- "Storage scheduler" local to each worker which keeps all storage objects within capacity limits
- Some sort of redirection for a `Chunk` whose data is potentially no longer stored directly in memory, but is stored somewhere else. This lets us not have to update `Chunk` arguments for thunks we haven't yet seen in the Sch/storage scheduler.
- Move `CHUNK_CACHE` logic into storage scheduler, and let it track all local versions of a `Chunk`'s data
- Storage scheduler can evict data from any storage location and store it somewhere else, and even evict it entirely if it's not owned by the local worker
- The owner of a `Chunk` always keeps at least a single copy of the data around locally
- Some way to represent storage objects with overlapping backing devices (like when we have two storage objects saving things in different directories but on the same filesystem)
[...]

Issue #183, opened 2020-12-26
Checkpointing infrastructure


jpsamaroo:
If at some point during a Dagger computation a thunk throws an error, or if the entire computation dies because the head node hit an OOM or other unexpected error, the entire computation is lost and needs to be started from scratch. This can be unacceptable for scheduling very large/expensive/mission-critical graphs, and for interactive development where errors are common and easily fixable.

Robust applications often support "checkpointing", where intermediate results are periodically written out to persistent media, or sharded to the rest of the cluster, to allow resuming an interrupted computation from a point later than the original start. Dagger should provide infrastructure to perform user-driven (or even automatic) checkpointing of intermediate results once they're generated, with the checkpoint/resume implementation being user-specified (although a default could be serializing to a temporary file with Julia's serializer).

PR #230, opened 2021-07-16
DTable implementation 

krynju:
TODO:

- [x] Tables.jl compatible constructors 
- [x] Constructor for prepartitioned data (proto for now, needs the planned storage api for full implementation)
- [ ] fetch(d::DTable) into any Tables.jl data structure (currently back to DataFrame)
- [ ] Implement the Tables.jl interface for the DTable (+ figure out what makes sense to implement and what doesn't)
- [ ] add more TODO items

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,6 +15,8 @@ SharedArrays = "1a1011a3-84de-559e-8e89-a11a2f7dc383"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+TableOperations = "ab02a1b2-a7df-11e8-156e-fb1833f50b87"
+Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
 Colors = "0.10, 0.11, 0.12"
@@ -24,7 +26,10 @@ StatsBase = "0.28, 0.29, 0.30, 0.31, 0.32, 0.33"
 julia = "1.0"
 
 [extras]
+Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
+CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test"]
+test = ["Test", "Arrow", "CSV", "DataFrames"]
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Dagger"
 uuid = "d58978e5-989f-55fb-8d15-ea34adc7bf54"
-version = "0.11.10"
+version = "0.12.0"
 
 [deps]
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
--- a/src/Dagger.jl
+++ b/src/Dagger.jl
@@ -45,7 +45,7 @@ include("array/sparse_partition.jl")
 include("array/sort.jl")
 
 include("ui/graph.jl")
-
+include("table/dtable.jl")
 function __init__()
     @require Luxor="ae8d54c2-7ccd-5906-9d76-62fc9837b5bc" begin
         # Gantt chart renderer
--- a/src/table/dtable.jl
+++ b/src/table/dtable.jl
@@ -1,0 +1,124 @@
+import Tables
+import TableOperations
+
+import Base: fetch
+
+export DTable, tabletype
+
+const VTYPE = Vector{Union{Dagger.Chunk,Dagger.EagerThunk}}
+
+"""
+    DTable
+
+Structure representing the distributed table based on Dagger.
+
+The table is stored as a vector of `Chunk` structures which hold partitions of the table.
+That vector can also store `EagerThunk` structures when an operation that modifies
+the underlying partitions was applied to it (currently only `filter`).
+"""
+mutable struct DTable
+    chunks::VTYPE
+    tabletype
+    DTable(chunks::VTYPE, tabletype) = new(chunks, tabletype)
+end
+
+DTable(chunks::Vector{Dagger.EagerThunk}, args...) = DTable(VTYPE(chunks), args...)
+DTable(chunks::Vector{Dagger.Chunk}, args...) = DTable(VTYPE(chunks), args...)
+
+include("iterators.jl")
+include("operations.jl")
+
+"""
+    DTable(table, chunksize; tabletype=nothing) -> DTable
+
+Constructs a `DTable` using a `Tables.jl` compatible `table` input.
+It assumes no initial partitioning of the table and uses the `chunksize`
+argument to partition the table (based on row count).
+
+Providing `tabletype` kwarg overrides the internal table partition type.
+"""
+function DTable(table, chunksize::Integer; tabletype=nothing)
+    if !Tables.istable(table)
+        throw(ArgumentError("Provided input is not Tables.jl compatible."))
+    end
+
+    parts = Tables.partitions(TableOperations.makepartitions(table, chunksize))
+    sink = Tables.materializer(tabletype !== nothing ? tabletype() : table)
+
+    chunks = Vector{Dagger.Chunk}()
+    sizehint!(chunks, length(parts))
+
+    type = nothing
+
+    for p in parts
+        tpart = sink(p)
+        push!(chunks, Dagger.tochunk(tpart))
+        if type === nothing
+            type = typeof(tpart).name.wrapper
+        end
+    end
+
+    return DTable(chunks, type)
+end
+
+"""
+    DTable(loader_function, files::Vector{String}; tabletype=nothing)
+
+Constructs a `DTable` using a list of filenames and a `loader_function`.
+Partitioning is based on the contents of the files provided, which means that
+one file is used to create one partition.
+
+Providing `tabletype` kwarg overrides the internal table partition type.
+"""
+function DTable(loader_function, files::Vector{String}; tabletype=nothing)
+    chunks = Vector{Dagger.EagerThunk}()
+    sizehint!(chunks, length(files))
+
+    append!(chunks, map(file -> Dagger.spawn(_file_load, file, loader_function, tabletype), files))
+
+    return DTable(chunks, tabletype)
+end
+
+function _file_load(filename, loader_function, tabletype)
+    part = loader_function(filename)
+    sink = Tables.materializer(tabletype === nothing ? part : tabletype())
+    tpart = sink(part)
+    return tpart
+end
+
+"""
+    fetch(d::DTable)
+
+Collects all the chunks in the `DTable` into a single, non-distributed
+instance of the underlying table type.
+"""
+function fetch(d::DTable)
+    sink = Tables.materializer(tabletype(d)())
+    sink(_retrieve_partitions(d))
+end
+
+"""
+    fetch(d::DTable, sink)
+
+Collects all the chunks in the `DTable` into a single, non-distributed
+instance of table type created using the provided `sink` function.
+"""
+fetch(d::DTable, sink) = sink(_retrieve_partitions(d))
+
+_retrieve_partitions(d::DTable) = TableOperations.joinpartitions(Tables.partitioner(_retrieve, d.chunks))
+
+_retrieve(x::Dagger.EagerThunk) = fetch(x)
+_retrieve(x::Dagger.Chunk) = collect(x)
+
+"""
+    tabletype(d::DTable)
+
+Provides the type of the underlying table partition.
+"""
+function tabletype(d::DTable)
+    if d.tabletype === nothing
+        f = c -> typeof(c).name.wrapper
+        d.tabletype = fetch(Dagger.@spawn f(d.chunks[1]))
+    end
+    return d.tabletype
+end
--- a/src/table/operations.jl
+++ b/src/table/operations.jl
@@ -1,0 +1,63 @@
+import Base: filter, map, reduce
+
+"""
+    map(f, d::DTableRows)
+
+Applies `f` to each row of `d`.
+
+# Examples
+```
+    d = DTable((a=rand(1000), b=rand(1000)), chunksize=100)
+    m = map(x -> x.a + x.b, eachrow(d))
+    fetch(m)
+```
+"""
+function map(f, d::DTableRows)
+    chunk_wrap = (chunk, f) -> begin
+        res = chunk |> TableOperations.map(x-> (result=f(x),)) |> Tables.materializer(chunk)
+        Tables.getcolumn(res, :result)
+    end
+    result_vector = map(c -> Dagger.spawn(chunk_wrap, c, f), d.dtable.chunks)
+    Dagger.@spawn vcat(result_vector...)
+end
+
+"""
+    reduce(f_reduce::Function, f_row::Function, d::DTableRows; init)
+
+Reduces `d` using function `f_reduce` on values obtained from applying `f_row` on its rows.
+
+# Examples
+```
+    d = DTable((a=rand(1000), b=rand(1000)), chunksize=100)
+    r = reduce(+, row -> row.a + row.b, eachrow(d); init=0.0)
+    fetch(r)
+```
+"""
+function reduce(f_reduce::Function, f_row::Function, d::DTableRows; init)
+    chunk_results = map(c -> Dagger.spawn(_chunk_reduce, f_reduce, f_row, c, init), d.dtable.chunks)
+    reduce_chunks = (f, _init, c_res...) -> reduce(f, c_res; init=_init)
+    Dagger.@spawn reduce_chunks(f_reduce, init, chunk_results...)
+end
+
+function _chunk_reduce(f_reduce, f_row, chunk, init)
+    mapped = chunk |> TableOperations.map(x -> (result = f_row(x)))
+    reduce(f_reduce, mapped; init=init)
+end
+
+"""
+    filter(f, d::DTable)
+
+Filter `d` using `f`.
+Returns a filtered `DTable` that can be processed further.
+
+# Examples
+```
+    d = DTable((a=rand(1000), b=rand(1000)), chunksize=100)
+    f = filter(x -> x.a > 0.5, eachrow(d))
+    fetch(f)
+```
+"""
+function filter(f, d::DTable)
+    chunk_wrap = (chunk, f) -> TableOperations.filter(f, chunk) |> Tables.materializer(chunk)
+    DTable(map(c -> Dagger.spawn(chunk_wrap, c, f), d.chunks), d.tabletype)
+end
--- a/test/table.jl
+++ b/test/table.jl
@@ -1,0 +1,134 @@
+using DataFrames
+using Arrow
+using CSV
+
+@testset "dtable" begin
+    @testset "constructors - Tables.jl compatibility (NamedTuple)" begin
+        size = 1_000
+        nt = (a = rand(Int, size), b = rand(Int, size))
+
+        dt = DTable(nt, 100)
+        @test fetch(dt) == nt
+        @test fetch(dt, NamedTuple) == nt
+        @test tabletype(dt) == NamedTuple
+    end
+
+    @testset "constructors - Tables.jl compatibility (DataFrames)" begin
+        size = 1_000
+        nt = (a = rand(Int, size), b = rand(Int, size))
+        df = DataFrame(nt)
+
+        dt = DTable(df, 100)
+        @test fetch(dt) == df
+        @test fetch(dt, DataFrame) == df
+        @test tabletype(dt) == DataFrame
+
+        dt = DTable(nt, 100; tabletype=DataFrame)
+        @test fetch(dt) == df
+        @test fetch(dt, DataFrame) == df
+        @test tabletype(dt) == DataFrame
+
+        dt = DTable(df, 100; tabletype=NamedTuple)
+        @test fetch(dt) == nt
+        @test fetch(dt, DataFrame) == df
+        @test tabletype(dt) == NamedTuple
+    end
+
+    @testset "constructors - Tables.jl compatibility (Arrow)" begin
+        size = 1_000
+        nt = (a = rand(Int, size), b = rand(Int, size))
+        io = IOBuffer()
+        Arrow.write(io, nt)
+        arr = Arrow.Table(take!(io))
+
+        dt = DTable(arr, 100)
+        @test fetch(dt) == nt
+        @test fetch(dt, NamedTuple) == nt
+        @test tabletype(dt) == NamedTuple # NamedTuple is the type obtained using the sink function of Arrow.Table
+    end
+
+    @testset "constructors - Tables.jl compatibility (CSV)" begin
+        size = 1_000
+        nt = (a = rand(Int, size), b = rand(Int, size))
+        io = IOBuffer()
+        CSV.write(io, nt)
+        dt1 = CSV.read(take!(io), (csv -> DTable(csv, 100)))
+        CSV.write(io, nt)
+        # or
+        dt2 = DTable(CSV.File(take!(io)), 100)
+        @test fetch(dt1) == nt
+        @test fetch(dt1, NamedTuple) == nt
+        @test tabletype(dt1) == NamedTuple
+        @test fetch(dt2) == nt
+        @test fetch(dt2, NamedTuple) == nt
+        @test tabletype(dt2) == NamedTuple
+    end
+
+    @testset "constructors - file input" begin
+        n = 20
+        size = 1000
+        ios = [IOBuffer() for _ in 1:n]
+        data = [(a = rand(size), b = rand(size)) for _ in 1:n]
+        arr = [Arrow.write(ios[idx], data[idx]) for idx in 1:n]
+
+        da = DTable(x -> Arrow.Table(take!(ios[tryparse(Int64, x)])), [string(i) for i in 1:n])
+        db = vcat([DataFrame(d) for d in data]...)
+        @test fetch(da, DataFrame) == db
+    end
+
+    @testset "map" begin
+        size = 1_000
+        nt = (a = rand(size), b = rand(size))
+        s = DataFrame(nt)
+        d = DTable(s, 100)
+        sm = map(x -> x.a + x.b, eachrow(s))
+        dm = map(x -> x.a + x.b, eachrow(d))
+        @test fetch(dm) == sm 
+        d = DTable(nt, 100)
+        dm = map(x -> x.a + x.b, eachrow(d))
+        @test fetch(dm) == sm
+    end
+
+    @testset "filter" begin
+        size = 1_000
+        nt = (a = rand(size), b = rand(size))
+        df = DataFrame(nt)
+        dt = DTable(df, 100)
+        dfr = filter(x -> x.a > 0.5, df)
+        dtr = filter(x -> x.a > 0.5, dt)
+        @test fetch(dtr) == dfr
+        dt = DTable(nt, 100)
+        dtr = filter(x -> x.a > 0.5, dt)
+        @test fetch(dtr, DataFrame) == dfr
+    end
+
+    @testset "reduce" begin
+        size = 1_000
+        nt = (a=rand(Int, size) .% 1000, b=rand(Int, size) .% 1000)
+
+        s = DataFrame(nt)
+        d = DTable(s, 100)
+        dn = DTable(nt, 100)
+
+        dr1 = reduce(+, y -> y.a, eachrow(d);init=0)
+        dn1 = reduce(+, y -> y.a, eachrow(dn);init=0)
+        sr1 = reduce((x, y) -> x + y.a, eachrow(s);init=0)
+
+        dr2 = reduce(+, y -> y.a * y.b, eachrow(d);init=0)
+        dn2 = reduce(+, y -> y.a * y.b, eachrow(dn);init=0)
+        sr2 = reduce((x, y) -> x + y.a * y.b, eachrow(s);init=0)
+
+        dr3 = reduce(+, y -> y.a + y.b, eachrow(d);init=0)
+        dn3 = reduce(+, y -> y.a + y.b, eachrow(dn);init=0)
+        sr3 = reduce((x, y) -> x + y.a + y.b, eachrow(s);init=0)
+
+        dr4 = reduce(*, y -> y.a, eachrow(d);init=0)
+        dn4 = reduce(*, y -> y.a, eachrow(dn);init=0)
+        sr4 = reduce((x, y) -> x * y.a, eachrow(s);init=0)
+
+        @test fetch(dr1) == fetch(dn1) == sr1
+        @test fetch(dr2) == fetch(dn2) == sr2
+        @test fetch(dr3) == fetch(dn3) == sr3
+        @test fetch(dr4) == fetch(dn4) == sr4
+    end
+end
--- a/src/table/iterators.jl
+++ b/src/table/iterators.jl
@@ -1,0 +1,12 @@
+import Base: eachrow, eachcol
+
+struct DTableRows
+    dtable::DTable
+end
+
+struct DTableColumns
+    dtable::DTable
+end
+
+Base.eachrow(d::DTable) = DTableRows(d)
+Base.eachcol(d::DTable) = DTableColumns(d)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -31,6 +31,7 @@ include("checkpoint.jl")
 include("domain.jl")
 include("array.jl")
 include("cache.jl")
+include("table.jl")
 try # TODO: Fault tolerance is sometimes unreliable
 include("fault-tolerance.jl")
 catch


Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.11.3"
 
 [deps]
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
```
jpsamaroo, line 7: We should put the DataFrames-specific code behind an `@requires` block in `__init__()` so that we don't take on a DataFrames dependency. We should only need the Tables direct dependency.

jpsamaroo (changes requested):
Great work so far!

jpsamaroo:
@krynju when you get the chance, let's try benchmarking this against raw DataFrames and see how we do. I suspect we'll probably want some scheduler changes to optimize concurrent thunk submission (which currently takes a global scheduler lock).

ExpandingMan:
Hey guys, glad to see this exciting work, thanks for all the effort!

I feel a little awkward making such a radical suggestion as merely an interested observer, but are you sure it would not be better to split this into a new package dependent on Dagger, rather than relying on `@requires`?  This approach has worked out quite well in the JuliaData ecosystem where many packages have delightfully minimal dependencies: for example several extensions to DataFrames.jl exist outside that package such as DataFramesMeta.jl, IO packages such as CSV.jl which despite being used with DataFrames.jl in the majority of use cases only rely on the Tables.jl interface.  There are even alternative tables packages such as TypedTables.jl which implement the Tables interface and are completely independent in terms of dependencies.  If I make any significant contribution to the distributed tables stuff myself, it will likely be in expanding the functionality of Parquet.jl to be compatible with your work here, and it's not yet clear to me whether this will involve dependencies on DTables at all, perhaps it will only need Dagger and perhaps it will only need Tables.

I realize that currently this isn't a lot of code, so splitting it may seem premature, but it could easily grow over time to be an enormous code base, so splitting it can make dependency logic much more straightforward, and it is often easier to split projects in their nascent stages rather than having a massive reorganization later on.

src/table/dtable.jl
```diff
@@ -0,0 +1,46 @@
+import DataFrames
+import Tables
+
+import Base: fetch, filter, map, setproperty!, eachrow
+
+mutable struct DTable
+    v::Vector{Dagger.EagerThunk}
+    thunksize::Int
```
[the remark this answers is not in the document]
krynju: I agree, I've added this at some point and didn't really think about it.
The chunk/partition size would only be relevant on creation and only in case the source needs to be partitioned by the constructor.
At this point I'll remove it as I don't really see a point in keeping it either

src/table/dtable.jl
```diff
@@ -0,0 +1,46 @@
...
+end
+
+function DTable(table::DataFrames.DataFrame; thunksize=10)
```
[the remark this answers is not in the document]
krynju: Yes, Tables.jl input and making the DTable Tables.jl compatible is definitely on the todo list.
Tried to do it from the start, but between DataFrames and Arrow I had some issues with different behavior so I just left it on DataFrames for now.

src/table/dtable.jl
```diff
@@ -0,0 +1,46 @@
...
+    if (!Tables.istable(table))
+        throw("Provided input is not Tables.jl compatible.")
+    end
+    r = Dagger.@spawn Tables.rows(table)
+    n = DataFrames.nrow(table)
+    partition_rows = (x, l, r) -> getindex(x, l:r)
+    df_create = (rows,i) -> DataFrames.DataFrame( partition_rows(rows, i, (i + thunksize - 1) % (n + 1)))
+    return DTable([Dagger.@spawn df_create(r,i) for i in 1:thunksize:n], thunksize)
+end
```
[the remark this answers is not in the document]
krynju: So currently if you run fetch() on a DTable (either just created or one that had filter ran on it) it will vcat the chunks back into a DataFrame

jpsamaroo:
I'm fine with moving this functionality out once it's ready to go; in the meantime we can keep it here for easy access to CI, and for testing in case we need/want to make scheduler changes as part of this work.

xiaodaigh:
Yeah. I think Dagger can exist without a table so I think the table functionality should be moved out as well. 

Now for practicality reason, perhaps it's ok to keep it here for CI reason etc.

Project.toml
krynju: I've put it in the @requires block, but then the exported functions and structures are not available after importing the package into env.  I'll check out later how to handle this

vchuravy:
On alternative is to make it a subpackage, slightly annoying CI setup but makes joined development easier without bringing in unecessary dependencies or using Requires.

src/table/dtable.jl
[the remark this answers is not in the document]
krynju: Added a Tables.jl compatible constructor + tests for CSV, Arrow, DataFrames

krynju:
@jpsamaroo I've been thinking about managing the chunks that contain the partitions in a scenario where the table would be bigger than the available memory. 
So at this stage this implementation doesn't care and loads all the partitions into memory, so naturally the process will die when it goes above the limit (at least when using threads, didn't check with processes)

So my idea for now is to first adjust the chunksize to memory size instead of rowcount to enable the rest of the functionality.
Then there's two scenarios:
1. Tables loaded using the Tables.jl interface - the table will need to be partitioned within the constructor according to the chunksize. There will have to be a mechanism of caching some chunks into disk cache and then loading them back on demand.
2. Tables that were initialized with a set of files - these will initially be partitioned according to the files provided and loaded into memory on demand. Then in case of any repartitioning (grouping, sorting etc.) the table will have to be treated as the table from point 1, so it will be partitioned and cached to the disk normally.

So both of these assume we would have some form of visibility on the memory available for a Processor to figure out when to cache the chunks and when not to load them into memory. I would probably declare/infer that information on the OSProc level, so even in a Distributed environment that memory availability could be handled. A different problem is where to cache the data - local env is easy, but in Distributed we should probably cache the table chunks at some commonly available location.

I'm wondering how much of this should be a part of the Dagger internals/scheduler and what should be a part of the DTable. I could imagine having a CachableChunk type for which you could define cache/restore functions and the scheduler figuring out when to call the cache and restore on these in order to make space in the physical memory and prepare the chunks for the Tasks to execute. With something like that the DTable wouldn't have a lot of work to do to provide out of core processing.

What's your take on this?

jpsamaroo:
I've thought about swap-to-disk a lot recently; some initial thoughts are here: https://github.com/JuliaParallel/Dagger.jl/issues/214#issuecomment-847926781, which sound quite similar to what you're suggesting.

> I'm wondering how much of this should be a part of the Dagger internals/scheduler and what should be a part of the DTable.

This is definitely a thing that the scheduler should know about and handle automatically. It's also relevant to the DArray, and generally to other kinds of operations that generate lots of data.

> I could imagine having a CachableChunk type for which you could define cache/restore functions and the scheduler figuring out when to call the cache and restore on these in order to make space in the physical memory and prepare the chunks for the Tasks to execute.

We already have the cache/restore in the form of checkpointing available to thunks (https://github.com/JuliaParallel/Dagger.jl/issues/183), although this is more of an automatic operation, whereas checkpointing is more manual (although that can change). Regardless, I don't think we need a new chunk type; the existing `Chunk` has a `handle` field that can be a `MemPool.FileRef`, or something else of our construction.

> With something like that the DTable wouldn't have a lot of work to do to provide out of core processing.

It's my hope that we can handle all of these details in the scheduler, so that any Dagger computation can benefit from swap-to-disk. I don't think exposing it to the DTable would be necessary; any knobs could be controlled via multiple dispatch on the chunked data type.

krynju:
So essentially there's three separate topics:
1. Adding the scheduler memory awareness and management 
2. Chunk with a handle type specialized for loading parts of the table stored initially as a set of files reachable by some Processors. That would come with a user defined loader, so they can use any input format as long as at the end it produces a Tables.jl compatible source ingestible by the DataFrame.
3. Chunk with a handle type specialized for storing table partitions in memory with specialized cache/restore functionality, so that the Scheduler can manage these according to the memory limits on used Processors.

So 2 and 3 is separate, because in my mind a chunk from point 2 becomes a chunk from point 3 after being successfully loaded once (and in case the partition gets dropped accidentaly it can be loaded again from that source).

jpsamaroo:
> Chunk with a handle type specialized for loading parts of the table stored initially as a set of files reachable by some Processors. That would come with a user defined loader, so they can use any input format as long as at the end it produces a Tables.jl compatible source ingestible by the DataFrame.

> Chunk with a handle type specialized for storing table partitions in memory with specialized cache/restore functionality, so that the Scheduler can manage these according to the memory limits on used Processors (other chunks would be just serialized/deserialized for memory management purposes, here we could optimize it for the partitions specifically).

I think we need that handle type (which can just be `MemPool.FileRef` potentially), and potentially also some way of expanding the `Processor` abstraction to reflect persistent storage. So the delineation would be:
- Chunk with handle of type `MemPool.DRef` (what's currently used to represent in-memory data), and processor that the data is associated with.
- Chunk with handle of type `MemPool.FileRef`, and "processor" representing the storage format of the data.

Here, the `DRef` and `FileRef` implement the means to retrieve stored data; doing `MemPool.poolget(::DRef)` returns the data stored in memory, while `MemPool.poolget(::FileRef)` should *ideally* return a representation of the data as stored on the file system (probably an open file descriptor on the host system). (Note that `MemPool.poolget` is called by `Dagger.move` by default). In the case of `FileRef`, it *should not* actually read the data into memory: that should be the job of `Dagger.move`, which will convert and serialize the data as needed. For a `FileRef`, doing a `Dagger.move` from some "storage processor" to a `ThreadProc` would imply reading the data into memory on that thread (really, just into the associated Julia process) in the most efficient manner possible. The combination of source processor, destination processor, and data type type should direct everything necessary to move data from file to memory, and back.

So what's missing here, in my mind, as an extra abstract type `Storage`, subtypes of which can be used as the `processor` field of `Chunk`, and represents that instead of having the data be associated with a computational device, it's associated with a storage device. It wouldn't be legal to execute code against a `Storage` type, so there'd be an extra step required to make a new `Chunk` with a real `Processor` (in other words, move it into memory).

> Adding the scheduler memory awareness and management

Given the above, the scheduler could be taught to move data from disk to memory when the data is necessary for a computation. It can also be taught that, when bringing in a new piece of data into memory in some way, if the memory requirement estimated for that data is greater than the memory limit of the processor (CPU RAM, GPU VRAM, etc.), then other chunks that're in-memory right now will need to first be moved to disk and cleared from memory.

An inability to perform this operation (say, because disk space ran out) would result in the scheduler setting an error state on that task of the DAG and anything that depends on it. In an interactive setting, the user could clean up more disk space *or* memory, and then re-run the computation and the scheduler should then retry from where it left off.

So what will be the format on disk for some arbitrary blob of data currently in-memory? This should probably be determined by a hypothetical `Dagger.default_storage_type(X)`, which would return a `Storage` subtype that can feasibly store `X`. Dagger would try to find a matching `Storage` instance, and if it succeeds, the scheduler could use that concrete instance to initiate a transfer into that storage location. The result of that transfer will be a `MemPool.FileRef`, which can later be used by the same `Storage` instance to retrieve the data from disk.


So how does all this work for the DTable? The DTable will need to auto-detect (based on the input table type or other mechanism) what `Storage` subtype is appropriate to represent the data is it is passed in, and then query Dagger to retrieve a matching storage instance. That can be used to construct any chunks associated with the input table, and those chunks can be safely passed as thunk arguments, and the scheduler will figure things out. Of course, the scheduler will then be free to temporarily send that data to disk in whatever format it wants (probably not CSV I would hope).


Thoughts?

krynju:
Makes sense to me with my current understanding of the internals

Although I have 3 questions:
1. Should the DTable storage stay as it is? Currently it's a `Vector{Dagger.EagerThunk}` and the thunk inside after creation is a simple `Dagger.@spawn (()->df)()`. If there's some operation applied the thunk will represent that operation instead. If I were to replace the storage with a `Vector{Chunk}` then I would have to unwrap the thunks with some table operations to retrieve the chunk and place it there, but I suppose that will break the execution graph if another operation were to be done on that retrieved chunk.
2. On creation (keeping in mind the new interface) I would have to explicitly create a chunk with .processor attribute set to the Storage instance the partition would be loaded from. What then? Should I wrap it in a simple thunk as well and keep it like that in the DTable vector? eg. `Dagger.@spawn collect(Dagger.tochunk(<partition>))`
3. If we ever wanted to give special treatment to the partitions caching/restore mechanism (maybe compression) how to specialize that? Implementing a `Dagger.move` operation for these chunks? How to figure out whether it is that partition chunk? New handle type?

shashi:
Is there any plans to support indexing? How would groupby and join work efficiently? I spoke to @quinnj  last year about possibly adding partitions and indexing helpers to Tables.. might to be a good time to restart that discussion.

jpsamaroo:
We might want to consider what's available with Tables and TableOperations now: https://youtu.be/Ryp0F0dHnLA

krynju:
I've seen that part of the interface and it will be very useful here both as a consumer of it and a source.

I've just added some small docs, fixed some consistency issues and put the DataFrames dependency in @requires. I think I can consider this initial scope ready as it is. 

Next todo's I think can be summarized by this list:

- [ ] Tables.jl (being a source) - figure out what parts of the interface and how to implement them in a way that makes sense.
- [ ] Tables.jl (consuming the interface) - add a constructor consuming the partitions interface
- [ ] Prepare a strategy on how to tackle repartitioning related operations: shuffle, sort, groupby, join; and assess their memory consumption.
- [ ] select and transform
- [ ] Implement the Storage API into the DTable when it's ready
- [ ] Figure out indexing
- [ ] Figure out potential schema usage/discovery and whether it can be any useful or just keep it unknown
- [ ] Add some convenience functions to inspect the contents of the DTable such as getindex, getcolumn, first, last, describe, schema information, length, partition information, memory placement across Processors, cached percentage etc.
- [ ] Consider some form of DResult structure for returning results of operations first applied per chunk, but not yet aggregated. Useful in map if you don't want to immediately get a full vcatted vector on the master processor (not so much in reduce).

quinnj:
This is exciting! Feel free to ping me on Tables.jl review stuff; happy to help. Or even to brainstorm the implementation, happy to chat on slack or discourse.

src/table/operations.jl
```diff
@@ -0,0 +1,85 @@
...
+    d = DTable((a=rand(1000), b=rand(1000));chunksize=100)
+    m = map(x -> x.a + x.b, eachrow(d))
+    fetch(m)
+```
+"""
+function map(f, d::DTableRows)
+    chunk_wrap = chunk -> begin 
+        i = eachrow(chunk)
+        map(f, i)    
+    end
```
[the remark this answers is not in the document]
jpsamaroo: Although these really should be pulled out into dedicated functions, because otherwise we'll incur a lot of compile time (due to specializing the closure object on the captured `f`).

src/table/operations.jl
```diff
@@ -0,0 +1,85 @@
...
+        ast_f = Base.uncompressed_ast(methods(fun).ms[1])
+        return_line = ast_f.code[end]
+        if hasproperty(return_line, :val) 
+            ast_id_of_ret_val = return_line.val.id # julia 1.6+
+        else
+            ast_id_of_ret_val= return_line.args[1].id # julia 1.5
+        end
+        ast_agg_op = ast_f.code[ast_id_of_ret_val].args[1]
+        getfield(ast_agg_op.mod, ast_agg_op.name)
+    end
```
jpsamaroo, line 61: Yeah, this is no good. What's the reasoning for this?

src/table/operations.jl
```diff
@@ -0,0 +1,85 @@
...
+    #
+    
+    reduce_chunk_results = (a...) -> reduce(get_aggregate_fun(f), a; init)
+
+    Dagger.@spawn reduce_chunk_results(chunk_results...)
+end
+
+"""
+    function filter(f, d::DTable)
```
jpsamaroo, line 70:
```suggestion
    filter(f, d::DTable)
```

src/table/operations.jl
```diff
@@ -0,0 +1,85 @@
...
+
+Filter the `DTable` using `f`.
```
jpsamaroo, line 72:
```suggestion
Filter `d` using `f`.
```

src/table/operations.jl
```diff
@@ -0,0 +1,85 @@
...
+Returns a filtered DTable that can be processed further.
```
jpsamaroo, line 73:
```suggestion
Returns a filtered `DTable` that can be processed further.
```

jpsamaroo (changes requested):
Looking good, although we need to be careful about using closures, since they can add unacceptable compile time costs.

src/table/dresult.jl
```diff
@@ -0,0 +1,17 @@
+struct DResult
```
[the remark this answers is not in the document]
krynju: Oops, that's a prototype file I made for one of the ideas from the todo list, it's not referenced anywhere though, so I'll remove it for now.
Basically it's supposed to represent a result of an operation that is first applied per chunk and then aggregated using some function. For example: map is first applied per chunk and then result gets aggregated using vcat when you call fetch on it.
It would be useful if you never wanted to collect the result into memory and just save it to the disk, but apart from map I didn't really find any useful scenarios for usage of this

src/table/operations.jl
```diff
@@ -0,0 +1,85 @@
...
+    chunk_wrap = chunk -> begin 
+        i = eachrow(chunk)
+        map(f, i)    
+    end
+    result_vector = [Dagger.@spawn chunk_wrap(c) for c in d.dtable.chunks]
+    Dagger.@spawn ((r...) -> vcat((r)...))(result_vector...)
+end
+
+function map(f, d::DTableColumns)
+    throw("Not implemented")
```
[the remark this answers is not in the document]
krynju: I left it out intentionally for now. I don't ever do these kinds of operations, but I know it's a thing, so I just put a placeholder. Seems to me that implementation of this might get more complicated. I'll look into it at some point, but not in this PR

src/table/operations.jl
krynju: Well, if I assume the format of the user provided function to be 
`(acc, row) -> any_associative_fun(acc, any_user_transform(row))`
then I would want to figure out the `any_associative_fun` out of it to reduce the values obtained from reducing chunks.

I've been thinking about it all morning, but I think I've answered my question right as I was writing this answer.
I think instead of asking the user to provide a generic reduce function I'd rather ask them to provide `any_associative_fun` and `any_user_transform` separately. That would let me avoid this AST mess

I'll try that instead

src/table/dtable.jl
```diff
@@ -0,0 +1,116 @@
...
+    r = _retrieve.(d.chunks)
+    @assert all(isa.(r, DataFrames.DataFrame))
+    vcat(r...)
+end
+
+_retrieve(x::Dagger.EagerThunk) = fetch(x)
+_retrieve(x::Dagger.Chunk) = collect(x)
+
+function getcolumn(d::DTable, s::Symbol)
+    _f = (x) -> Dagger.@spawn getindex(x, :, s)
```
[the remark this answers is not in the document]
krynju: I'll check, but at some point somewhere it didn't work for me, so I went straight for getindex

src/table/operations.jl
```diff
@@ -0,0 +1,85 @@
...
+    fetch(m)
+```
+"""
+function map(f, d::DTableRows)
+    chunk_wrap = chunk -> begin 
+        i = eachrow(chunk)
+        map(f, i)    
+    end
+    result_vector = [Dagger.@spawn chunk_wrap(c) for c in d.dtable.chunks]
+    Dagger.@spawn ((r...) -> vcat((r)...))(result_vector...)
```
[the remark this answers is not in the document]
krynju: leftover mess from having Dagger.@spawn in there as well :smile:

src/table/dtable.jl
```diff
@@ -0,0 +1,116 @@
...
+
+Constructs a `DTable` using a `Tables.jl` compatible `table` input.
+It assumes no initial partitioning of the table and uses the `chunksize`
+keyword argument to partition the table (based on row count).
+"""
+function DTable(table; chunksize)
+    if !Tables.istable(table)
+        throw(ArgumentError("Provided input is not Tables.jl compatible."))
+    end
+    create_chunk = (rows) -> begin
```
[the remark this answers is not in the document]
krynju: Extracted it into separate a separate function into a file with all dataframes dependencies

src/table/dtable.jl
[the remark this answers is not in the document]
krynju: yeah, the macro doesn't capture x as a chunk/thunk argument, so it calls getindex on a chunk

```
julia> Dagger.getcolumn(d, :d)
ERROR: MethodError: no method matching getindex(::Dagger.Chunk{DataFrames.DataFrame, MemPool.DRef, OSProc}, ::Colon, ::Symbol)

julia> @macroexpand Dagger.@spawn d[:, :a]
:(d[:, :a])

julia> @macroexpand Dagger.@spawn getindex(d, :, :a)
quote
    #= /home/krynju/.julia/dev/Dagger/src/thunk.jl:275 =#
    var"#6#args" = (d, :, :a)
    #= /home/krynju/.julia/dev/Dagger/src/thunk.jl:276 =#
    (Dagger.spawn)(getindex, var"#6#args"...; )
end
```

src/table/dtable.jl
```diff
@@ -0,0 +1,116 @@
...
+```jldoctest
+    DTable(["a.csv", "b.csv"], CSV.File)
+    DTable(["a.arrow", "b.arrow"], Arrow.Table)
+```
+"""
+function DTable(files::Vector{String}, loader_function)
+    chunks = Vector{Dagger.Chunk}()
+    sizehint!(chunks, length(files))
+
+    _load = file -> loader_function(file)
```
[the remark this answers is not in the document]
krynju: Fixed, leftover from having @spawn there

src/Dagger.jl
```diff
@@ -65,6 +65,10 @@ function __init__()
             include("ui/video.jl")
         end
     end
+    @require DataFrames="a93c6f00-e57d-5684-b7b6-d8193f3e46c0" begin
+         # Distributed table 
+         include("table/dtable.jl")
```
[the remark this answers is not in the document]
krynju: Resolved

src/table/dresult.jl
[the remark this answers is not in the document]
krynju: Removed for now, topic is in todo

src/table/operations.jl
krynju: Replaced this monstrosity with an interface change to provide two separate functions

krynju:
@jpsamaroo applied requested changes
CI failing due to some unrelated issue on master

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+"""
+    struct DTable
+
+Structure representing the distributed table based on Dagger.
+
+The table is stored as a vector of `Chunk` structures which hold partitions of the table.
+That vector can also store `EagerThunk` structures when an operation that modifies
+the underlying partitions was applied to it (currently only `filter`).
+
+Underlying partitions should always be `DataFrame` structures.
```
jpsamaroo, line 19:
```suggestion
    DTable

Structure representing the distributed table based on Dagger.

The table is stored as a vector of `Chunk` structures which hold partitions of the table.
That vector can also store `EagerThunk` structures when an operation that modifies
the underlying partitions was applied to it (currently only `filter`).
```

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
+import Tables
+import TableOperations
+
+import Base:fetch
```
jpsamaroo, line 4:
```suggestion
import Base: fetch
```

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+the underlying partitions was applied to it (currently only `filter`).
+
+Underlying partitions should always be `DataFrame` structures.
+"""
+struct DTable
+    chunks::VTYPE
+    tabletype
+    DTable(chunks::VTYPE, tabletype) = new(chunks, tabletype)
+    DTable(chunks::Vector{Dagger.EagerThunk}, args...; kwargs...) = new(VTYPE(chunks), args...; kwargs...)
+    DTable(chunks::Vector{Dagger.Chunk}, args...; kwargs...) = new(VTYPE(chunks), args...; kwargs...)
```
jpsamaroo, line 26:
```suggestion
    DTable(chunks::Vector{Dagger.EagerThunk}, args...; kwargs...) = DTable(VTYPE(chunks), args...; kwargs...)
    DTable(chunks::Vector{Dagger.Chunk}, args...; kwargs...) = DTable(VTYPE(chunks), args...; kwargs...)
```

Internal ctors don't take kwargs, so we need to forward to external ctors instead. Maybe move these out of the struct body?

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+include("operations.jl")
+
+""" 
+    DTable(table, chunksize) -> DTable
+
+Constructs a `DTable` using a `Tables.jl` compatible `table` input.
+It assumes no initial partitioning of the table and uses the `chunksize`
+argument to partition the table (based on row count).
+"""
+function DTable(table, chunksize::Integer;tabletype=nothing)
```
jpsamaroo, line 39:
```suggestion
function DTable(table, chunksize::Integer; tabletype=nothing)
```

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+    if !Tables.istable(table)
+        throw(ArgumentError("Provided input is not Tables.jl compatible."))
+    end
+    parts = Tables.partitions(TableOperations.makepartitions(table, chunksize))
+    chunks = Vector{Dagger.Chunk}()
+    sizehint!(chunks, length(parts))
+    _transform = Tables.materializer(table)
+    if tabletype !== nothing
+        _transform = Tables.materializer(tabletype())
+    end
```
jpsamaroo, line 49:
```suggestion
    _transform = Tables.materializer(tabletype !== nothing ? tabletype() : table)
```

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+    append!(chunks, Dagger.tochunk.(_transform.(parts)))
```
jpsamaroo, line 50: This isn't ideal to do directly, but it's fine for now. Let's remember to use `Dagger.@spawn` for this in a future PR.

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+    return DTable(chunks, if tabletype === nothing typeof(table) else tabletype end)
```
jpsamaroo, line 51:
```suggestion
    return DTable(chunks, tabletype === nothing ? typeof(table) : tabletype)
```

It's mostly just preference, but I prefer a ternary expression when it's all on one line and very simple.

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+Partitioning is based on the contents of the files provided, which means that
+one file is used to create one partition.
+
+# Examples
+```jldoctest
+    DTable(CSV.File, ["a.csv", "b.csv"])
+    DTable(Arrow.Table, ["a.arrow", "b.arrow"])
+```
+"""
+function DTable(loader_function, files::Vector{String};tabletype=nothing)
```
jpsamaroo, line 67:
```suggestion
function DTable(loader_function, files::Vector{String}; tabletype=nothing)
```

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+    DTable(loader_function, files::Vector{String})
+
+Constructs a `DTable` using a list of filenames and a `loader_function`.
+Partitioning is based on the contents of the files provided, which means that
+one file is used to create one partition.
+
+# Examples
+```jldoctest
+    DTable(CSV.File, ["a.csv", "b.csv"])
+    DTable(Arrow.Table, ["a.arrow", "b.arrow"])
```
jpsamaroo, line 64: Won't these files need to exist for the doctest? If so, maybe we can generate and write them out in the doctest, and then read them back in?

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+function DTable(loader_function, files::Vector{String};tabletype=nothing)
+    chunks = Vector{Dagger.Chunk}()
+    types = Vector{DataType}()
+    sizehint!(chunks, length(files))
+    sizehint!(types, length(files))
+
+    _transform = Tables.materializer(loader_function(files[1]))
+    if tabletype !== nothing
+        _transform = Tables.materializer(tabletype())
+    end
```
jpsamaroo, line 76:
```suggestion
    _transform = tabletype !== nothing ? Tables.materializer(tabletype()) : Tables.materializer(loader_function(files[1]))
```

jpsamaroo: I also don't prefer the `files[1]` thing, but it's fine for now. In a future PR, we can do this lazily with `Dagger.@spawn` and figure it out in the task.

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+
+    for file in files
+        tpart = _transform(loader_function(file))
+        push!(chunks, Dagger.tochunk(tpart))
+        push!(types, typeof(tpart))
+    end
+
+    @assert all(x-> x == types[1], types)
```
jpsamaroo, line 84: In the future, let's plan to remove this requirement.

src/table/dtable.jl
```diff
@@ -0,0 +1,110 @@
...
+Collects all the chunks in the `DTable` into a single, non-distributed instance of the underlying table type.
+"""
+function fetch(d::DTable)
+    m = Tables.materializer(d.tabletype)
+    jp = TableOperations.joinpartitions(Tables.partitioner(_retrieve, d.chunks))
+end
+
+function fetch(d::DTable, sink)
+    sink(fetch(d))
+end
```
jpsamaroo, line 101:
```suggestion
fetch(d::DTable, sink) = sink(fetch(d))
```

src/table/operations.jl
```diff
@@ -0,0 +1,60 @@
...
+
+# Examples
+```jldoctest        
+    d = DTable((a=rand(1000), b=rand(1000)), chunksize=100)
+    m = map(x -> x.a + x.b, eachrow(d))
+    fetch(m)
+```
+"""
+function map(f, d::DTableRows)
+    chunk_wrap = chunk -> map(f, eachrow(chunk))
```
jpsamaroo, line 16: This should take `f` and `chunk` as arguments. I've found that closures like these will get specialized on `f`, which can cause excessive compilation when playing around in the REPL.

src/table/operations.jl
```diff
@@ -0,0 +1,60 @@
...
+
+# Examples
+```jldoctest        
+    d = DTable((a=rand(1000), b=rand(1000)), chunksize=100)
+    r = reduce(+, row -> row.a + row.b, eachrow(d); init=0.0)
+    fetch(r)
+```
+"""
+function reduce(f_reduce::Function, f_row::Function, d::DTableRows; init)
+    chunk_reduce = (agg, row) -> f_reduce(agg, f_row(row))
```
jpsamaroo, line 38: Should take `f_reduce` and `f_row` as arguments.

src/table/operations.jl
```diff
@@ -0,0 +1,60 @@
...
+    chunk_wrap = (chunk, f, init) -> reduce(f, eachrow(chunk); init=init)
+    chunk_results = [Dagger.@spawn chunk_wrap(c, chunk_reduce, init) for c in d.dtable.chunks]
+    reduce_chunks = (a...) -> reduce(f_reduce, a; init)
+    Dagger.@spawn reduce_chunks(chunk_results...)
```
jpsamaroo, line 42: Should take `f_reduce` and `init` as arguments.

src/table/operations.jl
```diff
@@ -0,0 +1,60 @@
...
+end
+
+"""
+    filter(f, d::DTable)
```
jpsamaroo, line 46:
```suggestion
    filter(f, d::DTable) -> DTable
```

src/table/operations.jl
```diff
@@ -0,0 +1,60 @@
...
+    result_vector = [Dagger.@spawn chunk_wrap(c) for c in d.dtable.chunks]
+    Dagger.@spawn vcat(result_vector...)
+end
+
+function map(f, d::DTableColumns)
+    throw("Not implemented")
+end
+
+"""
+    reduce(f_reduce::Function, f_row::Function, d::DTableRows; init)
```
jpsamaroo, line 26:
```suggestion
    reduce(f_reduce::Function, f_row::Function, d::DTableRows; init) -> DTable
```

src/table/operations.jl
```diff
@@ -0,0 +1,60 @@
+import Base: filter, map, reduce
+
+"""
+    map(f, d::DTableRows)
```
jpsamaroo, line 4:
```suggestion
    map(f, d::DTableRows) -> DTable
```

test/table.jl
```diff
@@ -0,0 +1,101 @@
+using Test
```
jpsamaroo, line 1:
```suggestion
```

test/table.jl
```diff
@@ -0,0 +1,101 @@
...
+        @test fetch(d) == DataFrame([(a = 10, b = 20) for i in 1:size])
+    end
+
+    @testset "constructors - Tables.jl compatibility (DataFrames)" begin
+        size = 1_000
+        sample_df = () -> DataFrame(a=rand(size), b=rand(size))
+        s = sample_df()
+        d = DTable(s, 100)
+        @test fetch(d) == s
+        end
```
jpsamaroo, line 19:
```suggestion
    end
```

test/table.jl
```diff
@@ -0,0 +1,101 @@
...
+    @testset "constructors - Tables.jl compatibility (Arrow)" begin
+        size = 1_000
+        io = IOBuffer()
+        Arrow.write(io, [(a = 10, b = 20) for i in 1:size])
+        t = Arrow.Table(take!(io))
+        d = DTable(t, 100)
+        @test fetch(d) == DataFrame(t)
+    end
+
+        @testset "constructors - Tables.jl compatibility (CSV)" begin
```
jpsamaroo, line 30:
```suggestion
    @testset "constructors - Tables.jl compatibility (CSV)" begin
```

test/table.jl
```diff
@@ -0,0 +1,101 @@
...
+        dr4 = reduce(*, y -> y.a, eachrow(d);init=0)
+        sr4 = reduce((x, y) -> x * y.a, eachrow(s);init=0)
+        
+        @test fetch(dr1) == sr1
+        @test fetch(dr2) == sr2
+        @test fetch(dr3) == sr3
+        @test fetch(dr4) == sr4
+    end
+end
+
```
jpsamaroo, line 101:
```suggestion
```

jpsamaroo (approved):
Great work! Just a few nitpicks and comments, and then I'd be happy to merge this! :smile:

src/table/dtable.jl
krynju: There would be two reasons to load it from the main process directly, so maybe there can be two separate constructors in the future for that.
1. Maybe only the main process has access to local directory/network location from which the data is being loaded
2. If there was ever a need to split files into partitions as well due to them being too big (without implementing the chunk vector as some kind of tree).

src/table/dtable.jl
krynju: It's out for now, but some form of check in some operations will definitely be needed in case of some operations requiring non-generic functionality

src/table/operations.jl
krynju: Doesn't return a DTable if that's the meaning

src/table/operations.jl
krynju: Doesn't return a DTable if that's the meaning

jpsamaroo:
Thanks a ton! I'm looking forward to seeing this develop :smile:

PR #234, opened 2021-07-17
Fix unsafe usage of Base.Condition

jpsamaroo:


jpsamaroo:
@krynju @ExpandingMan please give this a try

krynju:
Works fine for me! 
Also checked it with https://github.com/JuliaParallel/Dagger.jl/pull/223 merged and works well together

ExpandingMan:
Looks good to me as well, thanks for the quick fix.

Btw, I'm pretty sure this problem exists on your recent tag, so a new version tag is likely in order.

Also, this is making me nervous about tests, which don't seem to have been running.  Is your intention for master to always be passing?  We should probably figure out why tests aren't running.  If you'd like I can do a PR to run github workflow tests, though I'm not sure how nicely they play with multiple processes.

jpsamaroo:
> Btw, I'm pretty sure this problem exists on your recent tag, so a new version tag is likely in order.

I just submitted 0.11.5, if that's what you mean.

> Also, this is making me nervous about tests, which don't seem to have been running.

The Buildkite tests are the most important, and they're generally passing (with some stochastic failures on nightly that I don't have the bandwidth to debug right now). Windows is just very unreliable with Dagger testing, but it is good to have a sense for how well things are working for those users.

>  Is your intention for master to always be passing?

If possible, yes, but I can tolerate some failures due to random chance. Much of the unreliability in CI we see appears to be due to Distributed, which I haven't had time to sit down and figure out.

ExpandingMan:
Maybe this is a stupid question but are the buildkite tests visible somewhere?  I don't think I'm seeing them in the PR's...

jpsamaroo:
Oh, shoot, you're supposed to be able to see them. @DilumAluthge any ideas?

PR #253, opened 2021-08-07
Set Julia 1.3 as minimum supported version

jpsamaroo:
Once a new LTS is officially announced, 1.3 will also be dropped in favor of that release (probably 1.6).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,7 +26,7 @@ Requires = "1"
 StatsBase = "0.28, 0.29, 0.30, 0.31, 0.32, 0.33"
 Tables = "1.5"
 TableOperations = "1"
-julia = "1.0"
+julia = "1.3"
 
 [extras]
 Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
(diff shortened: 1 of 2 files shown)


krynju:
Ok I looked around at compats for the Tables and testing dependencies

Arrow 1.0 - tables interface was added in 0.3, but shortly after 1.0 was released, so 1.0 seems like a good target
CSV 0.8 - it dropped the DataFrames dependency at that release
DataFrames 1.0 - some Tables.jl interface changes were applied on 1.0, seems like a good target
Tables 1.1 - partition interface was added (used in dtable)
TableOperations 1.0 - makepartitions was added (used in dtable)

```
[compat]
Arrow = "1.0"
CSV = "0.8"
Colors = "0.10, 0.11, 0.12"
DataFrames = "1.0"
MemPool = "0.3.4"
Requires = "1"
StatsBase = "0.28, 0.29, 0.30, 0.31, 0.32, 0.33"
TableOperations = "1"
Tables = "1.1"
julia = "1.0"
```

jpsamaroo:
None of the CI failures seem to be due to this change.

PR #259, opened 2021-08-12
DTable docs

krynju:
Just some docs and a print/show for the DTable, so that it doesn't spam the screen with chunk prints


krynju:
I'm adding some interface consistency changes, so let's put this on hold

krynju:
Merged this into https://github.com/JuliaParallel/Dagger.jl/pull/265

PR #263, opened 2021-08-13
settable global eager SchedulerOptions

kolia:
There is currently no way to set `SchedulerOptions` for the eager scheduler. This makes the options a global ref and adds a setter/getter.

### DIFF
--- a/src/sch/eager.jl
+++ b/src/sch/eager.jl
@@ -22,7 +22,6 @@ otherwise the global eager scheduler options are set to
 function global_eager_scheduler_options(; kwargs...)
     if !isassigned(EAGER_SCH_OPTS) || !isempty(kwargs)
         EAGER_SCH_OPTS[] = SchedulerOptions(; kwargs...)
-        @show EAGER_SCH_OPTS[]
     end
     return EAGER_SCH_OPTS[]
 end
--- a/src/sch/eager.jl
+++ b/src/sch/eager.jl
@@ -19,7 +19,7 @@ otherwise the global eager scheduler options are set to
 # Returns
 - `SchedulerOptions`: The global options used for eager scheduling.
 """
-function global_eager_scheduler_options(; kwargs...)
+function eager_options(; kwargs...)
     if !isassigned(EAGER_SCH_OPTS) || !isempty(kwargs)
         EAGER_SCH_OPTS[] = SchedulerOptions(; kwargs...)
     end
--- a/src/sch/eager.jl
+++ b/src/sch/eager.jl
@@ -3,8 +3,42 @@ const EAGER_THUNK_CHAN = Channel(typemax(Int))
 const EAGER_ID_MAP = Dict{UInt64,Int}()
 const EAGER_CONTEXT = Ref{Context}()
 const EAGER_STATE = Ref{ComputeState}()
+const EAGER_SCH_OPTS = Ref{SchedulerOptions}(SchedulerOptions(;allow_errors=true))
 
 eager_context() = isassigned(EAGER_CONTEXT) ? EAGER_CONTEXT[] : nothing
+
+"""
+    global_eager_scheduler_options(; kwargs...)
+
+Set / retrieve the global eager scheduler options.
+If no `kwargs` are passed, the current `SchedulerOptions` is returned,
+otherwise the global eager scheduler options are set to
+`SchedulerOptions(;kwargs...)` and returned.
+# Keywords
+- `kwargs...`: SchedulerOptions keywords to be set
+# Returns
+- `SchedulerOptions`: The global options used for eager scheduling.
+"""
+function global_eager_scheduler_options(; kwargs...)
+    if !isassigned(EAGER_SCH_OPTS) || !isempty(kwargs)
+        EAGER_SCH_OPTS[] = SchedulerOptions(; kwargs...)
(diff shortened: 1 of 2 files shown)


src/sch/eager.jl
```diff
@@ -3,9 +3,43 @@ const EAGER_THUNK_CHAN = Channel(typemax(Int))
...
+Set / retrieve the global eager scheduler options.
+If no `kwargs` are passed, the current `SchedulerOptions` is returned,
+otherwise the global eager scheduler options are set to
+`SchedulerOptions(;kwargs...)` and returned.
+# Keywords
+- `kwargs...`: SchedulerOptions keywords to be set
+# Returns
+- `SchedulerOptions`: The global options used for eager scheduling.
+"""
+function global_eager_scheduler_options(; kwargs...)
```
jpsamaroo, line 22:
```suggestion
function eager_options(; kwargs...)
```

The eager API is already global (just one scheduler instance), so I feel like "global" is redundant here.

I'd be okay with keeping the "scheduler" part if you'd prefer; let me know what you think!

src/sch/eager.jl
```diff
@@ -3,9 +3,43 @@ const EAGER_THUNK_CHAN = Channel(typemax(Int))
...
+    if !isassigned(EAGER_SCH_OPTS) || !isempty(kwargs)
```
jpsamaroo, line 23: The `isassigned(EAGER_SCH_OPTS)` isn't really necessary since you already set it at module load time, right?

src/sch/eager.jl
```diff
@@ -3,9 +3,43 @@ const EAGER_THUNK_CHAN = Channel(typemax(Int))
...
+        EAGER_SCH_OPTS[] = SchedulerOptions(; kwargs...)
```
jpsamaroo, line 24: Instead of wholesale replacing it, maybe we could do an in-place modification? It feels annoying to have to specify everything again if you just want to set one option.

Thoughts?

src/sch/eager.jl
```diff
@@ -3,9 +3,43 @@ const EAGER_THUNK_CHAN = Channel(typemax(Int))
...
+        @show EAGER_SCH_OPTS[]
```
jpsamaroo, line 25:
```suggestion
```

jpsamaroo (changes requested):
Thanks for this PR! Aside from some nitpicks, just needs some tests and docs.

src/sch/eager.jl
kolia: I haven't fully grokked how the eager scheduling works yet, but if some thunks that got pushed to `EAGER_THUNK_CHAN` have not been processed yet, would a later call to the in-place version of `eager_options`  not change the scheduler options used for those thunks retroactively? Seems like previously pushed thunks should use whatever options were in effect when they were pushed.

src/sch/eager.jl
jpsamaroo: Note that you also need to update the docstrings.

src/sch/eager.jl
jpsamaroo: Basically, we use whatever options are set whenever we access the options struct. I've left it like this because I didn't want to have to implement the logic to support what you're proposing (and the tests to make sure it works correctly). If you want to implement your proposed semantics, feel free to file a PR; otherwise, go ahead and file an issue and I'll get to it at some point in the future :slightly_smiling_face:

src/sch/eager.jl
jpsamaroo: Note that you still need to update the docstrings and references to the old name.

src/sch/eager.jl
kolia: Oh ok, that's fine, the current semantics probably cover 99% of use-cases, so guess it's easiest to just stick with them.

kolia:
I tried to add some tests, but am realising I'm not sure how to get the scheduler options that was used when scheduling a thunk.

I've pushed a view of my attempts as a draft, and marked this PR as a draft. Is there a better way to get the relevant context / options?

kolia:
I also tried making `SchedulerOptions` mutable and making `eager_options` work in-place, but get the same errors in my draft tests. Not surprising since it's accessing the scheduler options at all that's not working.

jpsamaroo:
Fixes #308

src/sch/eager.jl
jpsamaroo: Sorry, I was entirely wrong about the in-place modification - `SchedulerOptions` is immutable, and I think it's reasonable for it to stay like that. Marking this as resolved.

jpsamaroo:
The reason why this isn't working is that we don't actually read `EAGER_SCH_OPTS` within `src/sch/Sch.jl` more than once; instead it's passed once, and never re-read. We'd just need to replace uses of the local variable `opts` with a read of `EAGER_SCH_OPTS[]`.

jpsamaroo:
@kolia since I can't seem to clone your branch locally (maybe the repo got deleted?), I'm going to open a new PR with your patches and fix things up. Feel free to update this PR before that if you're around and still interested :slightly_smiling_face:

jpsamaroo:
I'm closing this because, as noted in #308 , we'll be moving away from `SchedulerOptions` as a public API for the eager scheduler. Regardless, thanks for putting this together @kolia !

PR #265, opened 2021-08-14
DTable interface consistency and initial docs

krynju:
1. Made the map and reduce function act more like TableOperations and Base equivalents instead of trying to mimick DataFrames.
2. Map now returns a DTable
3. Reduce returns a NamedTuple with results of per-column reduction. Made it work nicely with init from Base and you can also select the columns for reduction in there
4. A page about DTable for the documentation - need to still work on it for better examples of usage.


krynju:
note to self: adjust indexing for 1.3 appveyor CI as v[begin] is not supported at 1.3

src/table/operations.jl
```diff
@@ -1,63 +1,127 @@
...
+        columns = isnothing(cols) ? Tables.columnnames(Tables.columns(_retrieve(d.chunks[1]))) : cols
+    else
+        return Dagger.@spawn NamedTuple()
+    end
 
-function _chunk_reduce(f_reduce, f_row, chunk, init)
-    mapped = chunk |> TableOperations.map(x -> (result = f_row(x)))
-    reduce(f_reduce, mapped; init=init)
+    chunk_reduce = (_f, _chunk, _cols, _init) -> begin
+        values = [reduce(_f, Tables.getcolumn(_chunk, c); init=deepcopy(_init)) for c in _cols]
```
krynju, line 79: @jpsamaroo Do you think the per-column-within-chunk reduction should be done within separate tasks as well? 
I think this could be a potential performance improvement with bigger chunks

jpsamaroo: We should definitely parallelize more rather than less, and parallelizing per-column within a chunk might give the scheduler better information about compute costs for reducing that column, since columns can have different types, and thus take more or less time to compute.

src/table/dtable.jl
```diff
@@ -118,7 +120,39 @@ Provides the type of the underlying table partition.
 function tabletype(d::DTable)
     if d.tabletype === nothing
         f = c -> typeof(c).name.wrapper
-        d.tabletype = fetch(Dagger.@spawn f(d.chunks[1]))
+        if length(d.chunks) > 0
+            d.tabletype = fetch(Dagger.@spawn f(d.chunks[1]))
+        else # fallback
```
krynju, line 125: todo document this fallback

src/table/operations.jl
```diff
@@ -1,63 +1,127 @@
...
+        (; zip(_cols, values)...)
+    end
+    chunk_reduce_results = [Dagger.@spawn chunk_reduce(f, c, columns, deepcopy(init))  for c in d.chunks]
+
+    construct_single_column = (_col, _chunk_results...) -> getindex.(_chunk_results, _col)
+    result_columns = [Dagger.@spawn construct_single_column(c, chunk_reduce_results...) for c in columns]
+
+    reduce_result_column = (_f, _c, _init) -> reduce(_f, _c; init=_init)
+    reduce_chunks = [Dagger.@spawn reduce_result_column(f, c, deepcopy(init)) for c in result_columns]
```
krynju, line 88: So this part first takes the per chunk results and makes columns out of them and then reduces them.
I tried treereduce instead of this and it was noticeably slower.
I think for the DTable where we know there's not going to be more than <reasonable number> of chunks this could potentially be always faster than treereduce
Is there any case to use treereduce instead? Actually multimachine distributed maybe?

jpsamaroo (approved):
Awesome work!

src/table/operations.jl
krynju: I've added the spawns per column inside and:

1. Performance wise probably only an upgrade for many columns (need to look for a threshold sometime). For 2-4 cols it was usually a downgrade.
2. Stability wise it causes this https://github.com/JuliaParallel/Dagger.jl/issues/267 , so for now I'll keep this commented out

```julia
    col_in_chunk_reduce = (_f, _c, _init, _chunk) -> reduce(_f, Tables.getcolumn(_chunk, _c); init=deepcopy(_init))

    chunk_reduce = (_f, _chunk, _cols, _init) -> begin
        if length(_cols) <= 1 
            v = [col_in_chunk_reduce(_f, c, _init, _chunk) for c in _cols]
        else
            values = [Dagger.spawn(col_in_chunk_reduce, _f, c, _init, _chunk) for c in _cols]
            v = fetch.(values)
        end
        (; zip(_cols, v)...)
    end
```

src/table/dtable.jl
```diff
@@ -118,7 +120,39 @@ Provides the type of the underlying table partition.
 function tabletype(d::DTable)
```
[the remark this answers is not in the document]
krynju: Yeah makes sense. I'll do that with trim! as well, because that one happens on fetch which ideally shouldn't be mutating as well.

src/table/dtable.jl
```diff
@@ -103,56 +105,82 @@ instance of table type created using the provided `sink` function.
...
+
 """
     tabletype(d::DTable)
 
 Provides the type of the underlying table partition.
+
+In case the tabletype cannot be obtained the default return value is `NamedTuple`.
 """
 function tabletype(d::DTable)
-    if d.tabletype === nothing
```
jpsamaroo, line 119: Why remove this logic? Wouldn't we want to return the internal tabletype if one is already set? It would remove the need for the `_tabletype` logic I think.

test/table.jl
```diff
@@ -74,61 +78,137 @@ using CSV
...
-        d = DTable(s, 100)
-        sm = map(x -> x.a + x.b, eachrow(s))
-        dm = map(x -> x.a + x.b, eachrow(d))
-        @test fetch(dm) == sm 
-        d = DTable(nt, 100)
-        dm = map(x -> x.a + x.b, eachrow(d))
-        @test fetch(dm) == sm
+
+        df = DataFrame(nt)
+
```
jpsamaroo, line 92:
```suggestion

```

test/table.jl
```diff
@@ -74,61 +78,137 @@ using CSV
...
+            mt = map(x -> (r1 = x.a + x.b, r2 = x.a - x.b), dt)
+            mf = combine(df, [:a, :b] => ((a, b) -> a .+ b) => :r1, [:a, :b] => ((a, b) -> a .- b) => :r2)
+            @test fetch(mt).r1 == mf.r1
+            @test fetch(mt).r2 == mf.r2
+        end
+
+        # Map an empty dtable
+        dt = DTable((a = [], b = []), 10)
+        m = map(x -> (r = x.a + x.b), dt)
+        @test fetch(m) == NamedTuple()
```
jpsamaroo, line 111:
```suggestion
        @test fetch(m) == NamedTuple()
```

test/table.jl
```diff
@@ -74,61 +78,137 @@ using CSV
...
+
+            dfr = filter(x -> (x.a .> 0.5) .& (x.b .< 0.5), df)
+            dtr = filter(x -> (x.a .> 0.5) .& (x.b .< 0.5), dt)
+            @test fetch(dtr, DataFrame) == dfr
+        end
+
+        # Filter an empty DTable
+        dt = DTable((a = [], b = []), 10)
+        f = filter(x -> x.a .> 0.5, dt)
+        @test fetch(f) == NamedTuple()
```
jpsamaroo, line 131:
```suggestion
        @test fetch(f) == NamedTuple()
```

src/table/dtable.jl
krynju: Ok, I had the assumption in mind that there might be a case where you wouldn't want to use the cached tabletype, because for whatever reason it might be wrong, but now that I think about it there should be no case when it's actually wrong
I'll add the _tabletype logic to tabletype and remove _tabletype then.
I'll also make the tabletype! actually resolve the tabletype every time, because right now it would use the cached one to cache it.

jpsamaroo:
Thanks again!

Issue #150, opened 2020-10-13
Make it possible to add/remove processors at runtime


jpsamaroo:
The way processors are registered to Dagger is slightly ugly, and has the issue that we can't easily unregister processors later (because the callbacks in the list aren't uniquely identified). We should require identifiers be passed along with each callback so that it's trivial to remove them later (and we can unique on the name, which prevents double-registering the same processor).

Additionally, in the same vein as what #147 accomplished, we should make it possible to dynamically add and remove processors from a worker while the scheduler is executing. The user could then safely transition a computation across different groups of resources as their needs evolve.

PR #271, opened 2021-08-26
Wrap thunk functions in Chunk

### LINKED ISSUE #148: Cache serialized functions and closures
As mentioned in https://github.com/JuliaParallel/Dagger.jl/issues/115#issuecomment-703252872, we should perform the same kind of caching provided by `Distributed.CachingPool`.

### LINKED ISSUE #270: Treat thunk functions as Chunks
Currently we assume that thunk functions are immutable and cheap to serialize, but this isn't always the case. Flux, for example, uses callable structs to implement model layers, and in DaggerFlux.jl we end up accidentally serializing GPU arrays contained in the layers for every call (which works but is expensive because GPUArrays has a Serialization fallback).

We should consider handling non-`Function` thunk functions the same way we do for thunk arguments, by wrapping them in a `Chunk`. This would enable worker-local caching of the function to prevent excessive serialization, and by specifying a scope, prevent movement of mutable/non-serializable callable structs. In the case of DaggerFlux, layers would be assigned process-local scopes to prevent expensive serialization. To avoid under-utilization of cluster resources, models would need to be constructed within a thunk (and marked with a process-local scope) to ensure that workers will have local models to execute.

jpsamaroo:
Closes #270, closes #148 

TODO
- [x] Implement function awareness in `can_use_proc` and `schedule!`
- [x] Fix `Thunk` printing
- [x] Implement support in `delayed`
- [x] Handle function chunks in eviction
- [x] Fix any breakage in DaggerWebDash
- [x] Add basic tests for callable structs
- [x] Add tests for correct processor and scope
- [x] Add docstring for new kwargs
- [x] Add docs on using callable structs


jpsamaroo:
Now also partially addresses #150 by requiring a name (as `String` or `Symbol`) when adding processor callbacks.

PR #287, opened 2021-10-10
[DTable] Leftjoin

krynju:
- [x] leftjoin generic for any Tables.jl type
- [x] innerjoin generic for any Tables.jl type
- [x] inner matching passes with different kwargs for optimization (`r_sorted`, `l_sorted`, `r_unique`)
- [x] fallback to specialized methods (e.g. dataframes will use its own joins internally)
- [x] support for passing lookup tables to quickly match indices using a hash table for example
- [x] outerjoin research - answer: it's possible but not really nice of a solution, so skip for this PR
- [x] how to handle dtable x dtable joins? - answer: it works, but not necessairly the optimal way of doing it

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,6 +23,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 Colors = "0.10, 0.11, 0.12"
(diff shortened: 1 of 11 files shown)


jpsamaroo (approved):
Great work!

PR #296, opened 2021-10-20
catch_stack renamed current_exceptions in Julia 1.7

mattwigway:
`Base.catch_stack` was renamed `Base.current_exceptions` in Julia 1.7.

Tests pass on Julia 1.6 - I see some error backtraces in the console, but it ends with "Dagger tests passed," and it looks like some error backtraces are normal looking at the test run in CI. I also got a warning that parallel tests were running in serial. Tests are not passing yet on 1.7 - but they aren't on master either, and I don't think this change is related.

### DIFF
--- a/src/processor.jl
+++ b/src/processor.jl
@@ -174,7 +174,6 @@ iscompatible_arg(proc::ThreadProc, opts, x) = true
             fetch(task)
         catch err
             @static if VERSION >= v"1.1"
-                # catch_stack renamed current_exceptions in 1.7: https://github.com/JuliaLang/julia/pull/29901
                 @static if VERSION < v"1.7-rc1"
                     stk = Base.catch_stack(task)
                 else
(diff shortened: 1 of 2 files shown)


src/processor.jl
```diff
@@ -174,7 +174,12 @@ iscompatible_arg(proc::ThreadProc, opts, x) = true
             fetch(task)
         catch err
             @static if VERSION >= v"1.1"
-                stk = Base.catch_stack(task)
+                # catch_stack renamed current_exceptions in 1.7: https://github.com/JuliaLang/julia/pull/29901
```
jpsamaroo, line 177:
```suggestion
```

The code is pretty self-explanatory :slightly_smiling_face:

src/processor.jl
```diff
@@ -195,7 +200,12 @@ else
...
```
jpsamaroo, line 203:
```suggestion
```

src/sch/util.jl
```diff
@@ -194,7 +194,12 @@ function fetch_report(task)
         fetch(task)
     catch err
         @static if VERSION >= v"1.1"
-            stk = Base.catch_stack(task)
+            # catch_stack renamed current_exceptions in 1.7: https://github.com/JuliaLang/julia/pull/29901
```
jpsamaroo, line 197:
```suggestion
```

PR #304, opened 2021-11-09
DTable Tables.jl interface adjustments

krynju:
Applying adjustments from the code review on the original PR

OK, adjusted now. Removed the unnecessary definitions for Tables.jl (I don't want anyone to rely on them for anything) and added the ChainedVector to getcolumn, which requires an additional dependency.


src/table/tables.jl
```diff
@@ -45,17 +45,16 @@ function Tables.schema(table::DTable)
...
     s === nothing ? nothing : s.names
 end
 
 
 function _getcolumn(table::DTable, col::Union{Symbol, Int})
     chunk_col = (_chunk, _col) -> Tables.getcolumn(_chunk, _col)
     v = [Dagger.spawn(chunk_col, chunk, col) for chunk in table.chunks]
-    p = (_chunk) -> (; :_ => fetch(_chunk))
-    Tables.getcolumn(TableOperations.joinpartitions(Tables.partitioner(p, v)), :_)
+    SentinelArrays.ChainedVector(fetch.(v))
```
jpsamaroo, line 57: What's the purpose of using `ChainedVector` here?

jpsamaroo (approved):

src/table/tables.jl
krynju: It's to glue together the column vectors from all the chunks.
To be honest it would be best to use the `DArray` here or some equivalent lazy-loaded column representation, but this is good enough for now I guess

jpsamaroo:
Thanks!

PR #306, opened 2021-11-16
Add stub for DaggerFlux reverse CI

jpsamaroo:


jpsamaroo:
I'll add actual DaggerFlux reverse CI once we're sure that we can access the `juliagpu` BK queue, and https://github.com/JuliaGPU/DaggerGPU.jl/pull/17 is merged.

PR #311, opened 2021-12-05
Distribute the scheduler!

### LINKED ISSUE #165: Multithread and distribute the scheduler
The scheduler is quite close to being able to run multithreaded, with many parts running concurrently already. We're also already pushing some extra work onto the worker nodes (mainly argument fetching and processor selection/load balancing), and it would be beneficial for large DAGs to move more work onto each worker. The main blocker is providing a way to split the DAG into multiple domains, where each domain is handled by a given thread on a given worker. With efficient Thunk serialization, we can then send a subgraph to each worker and let them process their own DAG without conflicts. We'll need to add a mechanism by which thunks automatically wait on their input thunks to complete before they attempt to download the output data; if possible, we can also have workers broadcast and shard their `Chunk`s onto dependent workers as soon as the data is made available.

jpsamaroo:
This PR allows the scheduler to execute itself on all workers in the cluster. We first expand the notion of "thunk ID" to be per-worker, so that we can locally allocate unique IDs (and later locate where a thunk was created), and then allow the eager scheduler code to execute on all workers, instead of `remotecall`'ing to worker 1. We then allow thunks to be registered and scheduled locally (which should make recursive runtime-generated graphs vastly more efficient, no longer having to make a trip over the network). Finally, we implement local (and optionally remote?) work stealing (strictly for already-scheduled tasks, for the time being) to allow work to be kept balanced. The newly-available scheduler metrics on each worker will make it possible to optimize the choice of processor to steal from, although this can be left for later work.

Todo:
- [x] Execute the eager scheduler on all workers
- [ ] Implement local work stealing with ConcurrentCollection's `WorkStealingDeque`
- [ ] Tests for `@spawn`/`add_thunk!` with thunks owned by other schedulers
- [ ] Document new behavior
- [ ] Validate the web dashboard shows remote scheduler data
- [ ] Benchmarks

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,12 +4,12 @@
     queue: "juliaecosystem"
     sandbox.jl: "true"
 steps:
-  - label: Julia 1.6
+  - label: Julia 1.7
     timeout_in_minutes: 60
     <<: *test
     plugins:
       - JuliaCI/julia#v1:
-          version: "1.6"
+          version: "1.7"
       - JuliaCI/julia-test#v1:
           julia_args: "--threads=1"
       # - JuliaCI/julia-coverage#v1:
@@ -19,7 +19,7 @@ steps:
     <<: *test
     plugins:
       - JuliaCI/julia#v1:
-          version: "1.6-nightly"
+          version: "1.7-nightly"
       - JuliaCI/julia-test#v1:
           julia_args: "--threads=1"
       # - JuliaCI/julia-coverage#v1:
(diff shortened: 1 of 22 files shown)


jpsamaroo:
Closes https://github.com/JuliaParallel/Dagger.jl/issues/165

jpsamaroo:
I'm considering lower-bounding Julia to 1.7 going forward, so that we can easily work with atomics and use packages like ConcurrentCollections.jl and AtomicArrays.jl. If we do this, we'll do a minor version bump to 0.15, and keep 0.14.x as the last set of versions supporting Julia 1.6. We'll backport bug fixes and critical performance work to that branch if they're reproducible on Julia 1.6, and I'm also happy to backport any features that can be safely used on Julia 1.6 (assuming it doesn't rely on more recent changes, such as in this PR).

jpsamaroo:
This is being developed on `jps/dev`, and will be re-posted when ready.

PR #315, opened 2021-12-07
don't sort! parents when empty in order()

mattwigway:
This avoids the following error which crashes the scheduler:
```
MethodError: defalg(::Vector{Union{}}) is ambiguous. Candidates:
  defalg(v::AbstractArray{<:Union{Missing, Number}}) in Base.Sort at sort.jl:658
  defalg(::AbstractArray{<:Union{Missing, String1, String15, String3, String7}}) in InlineStrings at /Users/juliauser/.julia/packages/InlineStrings/C5tOh/src/InlineStrings.jl:656
Possible fix, define
  defalg(::AbstractArray{<:Missing})
```

When parents is empty in order(), it has type Vector{Union{}}. Since Union{} is a subtype of all
Union{...any types here...} types, any package that defines Base.defalg(AbstractArray{<:Union{...any type...}})
will cause an ambiguity error. This is true even without type piracy, as the Union in the other package could
consist entirely of types defined in that package.

@jpsamaroo

### DIFF
--- a/src/compute.jl
+++ b/src/compute.jl
@@ -166,7 +166,17 @@ function order(node::Thunk, ndeps)
         s += 1
         output[next] = s
         parents = collect(Iterators.filter(istask, inputs(next)))
-        sort!(parents, by=k->get(ndeps,k,0))
+        if !isempty(parents)
+            # If parents is empty, sort! should be a no-op, but raises an ambiguity error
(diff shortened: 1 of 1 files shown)


jpsamaroo (approved):
Thanks for patching this!

jpsamaroo:
Thanks!

PR #317, opened 2021-12-14
Make system UUID determination atomic

### LINKED ISSUE #316: Race in system_uuid_fallback
I just got this stacktrace while initializing Dagger on a large cluster:

<details>
<pre>
ERROR: LoadError: On worker 3:
InitError: ArgumentError: Malformed UUID string: ""
Stacktrace:
  [1] throw_malformed_uuid
    @ ./uuid.jl:80
  [2] parse
    @ ./uuid.jl:83 [inlined]
  [3] #154
    @ ~/.julia/packages/Dagger/xtTHY/src/utils/system_uuid.jl:13
  [4] get!
    @ ./dict.jl:464
  [5] system_uuid_fallback
    @ ~/.julia/packages/Dagger/xtTHY/src/utils/system_uuid.jl:4 [inlined]
  [6] system_uuid
    @ ~/.julia/packages/Dagger/xtTHY/src/utils/system_uuid.jl:20 [inlined]
  [7] __init__
    @ ~/.julia/packages/Dagger/xtTHY/src/Dagger.jl:69
  [8] _include_from_serialized
    @ ./loading.jl:768
  [9] _require_search_from_serialized
    @ ./loading.jl:854
 [10] _tryrequire_from_serialized
    @ ./loading.jl:783
 [11] _require_search_from_serialized
    @ ./loading.jl:843
 [12] _require
    @ ./loading.jl:1097
 [13] require
    @ ./loading.jl:1013
 [14] #1
    @ ~/dcm/julia-1.7.0/share/julia/stdlib/v1.7/Distributed/src/Distributed.jl:79
 [15] #103
    @ ~/dcm/julia-1.7.0/share/julia/stdlib/v1.7/Distributed/src/process_messages.jl:274
 [16] run_work_thunk
    @ ~/dcm/julia-1.7.0/share/julia/stdlib/v1.7/Distributed/src/process_messages.jl:63
 [17] run_work_thunk
    @ ~/dcm/julia-1.7.0/share/julia/stdlib/v1.7/Distributed/src/process_messages.jl:72
 [18] #96
    @ ./task.jl:423
during initialization of module Dagger
</pre>
</details>

I think the issue is [here](https://github.com/JuliaParallel/Dagger.jl/blob/master/src/utils/system_uuid.jl#L8). If multiple workers are running on the same machine, they may try to write to `dagger-system-uuid` simultaneously. Locking won't help us here, because the race is across processes.

Is there a reason we need to use UUIDs here? Could we just use the IP address of the worker machine?

jpsamaroo:
@mattwigway please review if you get the chance

Fixes #316

### DIFF
--- a/src/utils/system_uuid.jl
+++ b/src/utils/system_uuid.jl
@@ -2,16 +2,25 @@ const SYSTEM_UUIDS = Dict{Int,UUID}()
 
 function system_uuid_fallback()
     get!(SYSTEM_UUIDS, myid()) do
-        uuid_file = joinpath(tempdir(), "dagger-system-uuid")
+        username = get(ENV, "USER", "__global")
+        uuid_file = joinpath(tempdir(), "dagger-system-uuid-$username")
         if !isfile(uuid_file)
+            temp_uuid_file, temp_io = mktemp(; cleanup=false)
             uuid = uuid4()
-            open(uuid_file, "w") do io
-                write(io, string(uuid))
+            write(temp_io, string(uuid))
+            flush(temp_io)
+            close(temp_io)
+            try
+                # Try to make this the UUID file
+                mv(temp_uuid_file, uuid_file; force=false)
+                return uuid
+            catch err
+                err isa ArgumentError || rethrow(err)
+                # Failed, clean up temp file, and read existing UUID file
+                rm(temp_uuid_file)
             end
-            return uuid
-        else
-            return parse(UUID, read(uuid_file, String))
         end
+        return parse(UUID, read(uuid_file, String))
     end
     SYSTEM_UUIDS[myid()]
 end


mattwigway:
I think this should work, as `mv` on a single file system should in general be atomic. I don't think there would be any issue with multiple Dagger jobs sharing the same UUID per machine.

You mentioned in #316 that IP addresses can be the same between workers - but isn't the IP how `Distributed` knows how to connect to each worker, so shouldn't they be distinct? If we want to make sure things are unique per-machine, we could also use the MAC address of the first hardware network interface—although that may not work with Infiniband-fabric clusters like we have here at UNC, because the Infiniband hardware addresses are 20 bytes but are truncated when reading with `ifconfig`

jpsamaroo:
> I don't think there would be any issue with multiple Dagger jobs sharing the same UUID per machine.

I'd even consider that a feature, since in the future we might want to be able to bind disparate Julia clusters together, so having the UUID would let us find matching nodes across clusters.

> we could also use the MAC address of the first hardware network interface—although that may not work with Infiniband-fabric clusters like we have here at UNC, because the Infiniband hardware addresses are 20 bytes but are truncated when reading with ifconfig

MAC address would be a solid approach. Do we have an easy way to access that via libuv? If not, we might need to settle on IP addresses (although the approach I have here will generally work fine, since you don't typically share `/tmp` drives between networked machines).

Also, I don't want to be in the business of parsing script outputs, so using a library is the only reasonable way forward.

DrChainsaw:
>  but isn't the IP how Distributed knows how to connect to each worker, so shouldn't they be distinct?

I might be wrong here, but I think that port number is used as well so that the IP:portnr tuple must be distinct. One case when IPs are not distinct are clusters where a single machine can have multiple workers. Iirc I even had to replace the default port randomization as the birthday problem makes it so that large machines have a too large chance to fail due to port collisions.

mattwigway:
Isn't the point of this field to have a value that's unique to the physical
machine, not the worker process?

mattwigway:
Good point, we'd have that problem with MACs as well. The approach used in this PR is probably fine for now.

vchuravy (approved):

jpsamaroo:
Thanks all for the reviews!

> I am still not sure that Dagger needs to worry about this and instead we should use a network library like UCX that will do this for us.

Even if UCX can do optimized transfers via IPC, the scheduler still wants to know when those transfer types will be used, so that (in the future, at least) it can guesstimate their costs during scheduling.

PR #328, opened 2022-01-10
Add with_scope for propagating scopes

jpsamaroo:
Implements `Dagger.with_scope(f, scope)`, which sets the Dagger scope to `scope` while executing `f`. This allows us to do things like:

```julia
Dagger.with_scope(Dagger.ProcessScope(2)) do
  fetch(Dagger.@spawn 1+1)
end
```

which executes `1+1` on process 2. However, this feature goes further, supporting code like:

```julia
dt = DTable(...)
Dagger.with_scope(Dagger.ProcessScope(2)) do
  fetch(map(row->(;a=row.a+1), dt))
end
```

which will execute every task spawned by `map`, directly or indirectly, on process 2.

This is all implemented with ContextVariablesX.jl (@tkf) which makes it easy to propagate through nested tasks.

Todo:
- [ ] Tests, and lots of them!
- [ ] Document `with_scope` and `get_scope`
- [ ] Benchmark against master


tkf:
ContextVariablesX.jl is a playground for checking the "look and feel" of the API in the wild and so the implementation is rather... funny (it uses logger as the storage). It's definitely not as efficient as it can be and ultimately we need to put it in `Task`. I think ContextVariablesX.jl can still be improved somewhat though. Basically, we need some nice immutable or copy-on-write dictionary with small-size optimization. Currently, it copies `Dict` all the time and it could be a larger overhead than the choice of the storage backend.

jpsamaroo:
As long as it works correctly, I'm okay with a performance hit for now. I'm going to confirm that the impact is minimal in benchmarks, but otherwise I'd like to ship this before upstream develops official context variable support.

jpsamaroo:
Going forward, I think the approach I want to take is to enable Dagger to propagate arbitrary context variables through Dagger task spawns. This would allow us to propagate context variables from other libraries, which expands the possibilities for isolated execution contexts. We'll have to provide a way to opt-out for libraries whose context variables are intended to be hardware-local (such as the default GPU device, or some logger file path).

jpsamaroo:
I've updated this PR to work for any arbitrary set of options, which will make it possible for #289 to propagate an allocator/device.

PR #329, opened 2022-02-01
Sch: Start dynamic listener in init_proc

### LINKED ISSUE #325: dynamic_listener! not called for procs added after scheduler has started
Simple way to verify is to run only [these lines](https://github.com/JuliaParallel/Dagger.jl/blob/0e0379ba885e31afb9683d037af1027c79ff651a/test/scheduler.jl#L222-L244) of the add workers test, wait until everything is setup and then run the next line (test stalls as the new thunks are never started). I suppose that this might happen when running the test normally, but I have not seen it happen yet.

Changing the signature to `dynamic_listener!(ctx, state; tids = keys(state.worker_chans))` (and looping over `tids` in the main loop) and then calling it with only the new procs makes everything tick along as expected. 

I suppose the drawback is that there will be one task listening for halts each time this is done, so perhaps refactoring so that `dynamic_listener!` accepts a single proc and the `listener_tasks` is moved to state would be preferable. One should then call `dynamic_listener!` from `init_proc` as it is a required step for a proc to be operational.

I suppose one would also need/want a mechanism to clean up the listeners when procs are removed (although I dread dealing with the edge cases that this might create).

Let me know if you want a PR with any of the above options (or something else).

jpsamaroo:
Closes #325

@DrChainsaw

### DIFF
--- a/src/sch/Sch.jl
+++ b/src/sch/Sch.jl
@@ -286,6 +286,10 @@ function init_proc(state, p, log_sink)
     lock(state.lock) do
         state.worker_chans[p.pid] = (inp_chan, out_chan)
     end
+
+    # Setup dynamic listener
+    dynamic_listener!(ctx, state, p.pid)
+
     timespan_finish(ctx, :init_proc, p.pid, 0)
 end
 function _cleanup_proc(uid, log_sink)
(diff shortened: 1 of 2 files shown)


DrChainsaw:
Thanks alot for this and sorry for not doing it myself (I got bogged down in other stuff). I have not forgotten about revisiting the add/rm_proc! testcases and will try to do so whenever I find the time.

PR #332, opened 2022-02-22
Fix remove workers test

DrChainsaw:
As promised earlier, here is an attempt to fix the remove workers test.

It turned out a bit more evil than I had hoped as it was quite difficult to yank the workers after work had started and any alternative suggestions to using the logging framework for this is appreciated.

The change in `Sch` is because you'd otherwise get errors for not having ran `using Dagger` on the workers in the add workers test after it completes.

### DIFF
--- a/src/sch/Sch.jl
+++ b/src/sch/Sch.jl
@@ -821,7 +821,9 @@ function evict_all_chunks!(ctx, to_evict)
     end
 end
 function evict_chunks!(log_sink, chunks::Set{Chunk})
-    ctx = Context(;log_sink)
+    # Need worker id or else Context might use Processors which user does not want us to use.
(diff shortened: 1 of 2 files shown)


jpsamaroo:
Thanks! I'm going to let this sit for a day to hopefully get CI working again (since it seems like there are outages occurring across multiple tech platforms today), and hopefully get a review in, but the Sch changes LGTM!

DrChainsaw:
No problem.

Forgot to say above that don't hesitate to reject this if you find the usage of the logging framework too iffy. Could be to that the test is just too convoluted to be worthwhile. Not sure if there are any good ways to test things like this...

jpsamaroo:
Thanks!

PR #336, opened 2022-03-22
Add DArray documentation

mkitti:
This pull request adds `DArray` documentation.

### DIFF
--- a/docs/src/darray.md
+++ b/docs/src/darray.md
@@ -1,0 +1,204 @@
+# Distributed Arrays
+
+The `DArray`, or "distributed array", is an abstraction layer on top of Dagger
+that allows loading array-like structures into a distributed environment. The
+`DArray` partitions a larger array into smaller "blocks" or "chunks", and those
+blocks may be located on any worker in the cluster. The `DArray` uses a
+Parallel Global Address Space (aka "PGAS") model for storing partitions, which
+means that a `DArray` instance contains a reference to every partition in the
+greater array; this provides great flexibility in allowing Dagger to choose the
+most efficient way to distribute the array's blocks and operate on them in a
+heterogeneous manner.
+
+Aside: an alternative model, here termed the "MPI" model, is not yet supported,
+but would allow storing only a single partition of the array on each MPI rank
+in an MPI cluster. `DArray` support for this model is planned in the near
+future.
+
+This should not be confused with the [DistributedArrays.jl](https://github.com/JuliaParallel/DistributedArrays.jl) package.
+
+## Creating `DArrays`
+
+A `DArray` can be created in two ways: through an API similar to the usual
+`rand`, `ones`, etc. calls, or by distributing an existing array with
+`distribute`. Additionally, most operations on `DArray`s also return `DArray`s
+or an equivalent object which represents the operation being performed. It's
+generally not recommended to manually construct a `DArray` object unless you're
+developing the `DArray` itself.
+
+### Allocating new arrays
+
+As an example, one can allocate a random `DArray` by calling `rand` with a
+`Blocks` object as the first argument - `Blocks` specifies the size of
+partitions to be constructed. Note that the `DArray` is a lazy asynchronous
+object (i.e. operations on it may execute in the background), so to force it to
+be materialized, `fetch` may need to be called:
+
+```julia
+# Add some Julia workers
+julia> using Distributed; addprocs(6)
+6-element Vector{Int64}:
+ 2
+ 3
+ 4
+ 5
+ 6
+ 7
+
+julia> @everywhere using Dagger
+
+julia> DX = rand(Blocks(50, 50), 100, 100)
+Dagger.AllocateArray{Float64, 2}(100, 100)
+
+julia> fetch(DX)
+Dagger.DArray{Any, 2, typeof(cat)}(100, 100)
+```
+
+The `rand(Blocks(50, 50), 100, 100)` call specifies that a `DArray` matrix
+should be allocated which is in total 100 x 100, split into 4 blocks of size 50
+x 50, and initialized with random `Float64`s. Many other functions, like
+`randn`, `ones`, and `zeros` can be called in this same way.
+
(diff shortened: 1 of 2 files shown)


docs/src/darray.md
```diff
@@ -0,0 +1,135 @@
...
+Distributed arrays are created by invoking `compute` on an operation resulting in a distributed array.
+
+For example, one can allocate an array in a distributed fashion in the following fashion. To obtain a normal `Array`,
+`collect` can be used to gather the data from the various processes.
+```julia
+julia> DX = rand(Blocks(50,50), 100, 100)
+Dagger.AllocateArray{Float64, 2}(100, 100)
+
+julia> compute(DX)
+Dagger.DArray{Any, 2, typeof(cat)}(100, 100)
```
jpsamaroo, line 19: Following from the above comment, maybe we then say, "This call causes Dagger to actually allocate the randomly-initialized `DArray`. Without this call, the `DArray` is not yet materialized.".

docs/src/darray.md
```diff
@@ -0,0 +1,135 @@
...
+
+julia> DX = rand(Blocks(50,50), 100, 100)
+Dagger.AllocateArray{Float64, 2}(100, 100)
+
+julia> DY = DX + DX
+Dagger.BCast{Base.Broadcast.Broadcasted{Dagger.DaggerBroadcastStyle, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, typeof(+), Tuple{Dagger.AllocateArray{Float64, 2}, Dagger.AllocateArray{Float64, 2}}}, Float64, 2}(100, 100)
+
+julia> DZ = DY .* 3
+Dagger.BCast{Base.Broadcast.Broadcasted{Dagger.DaggerBroadcastStyle, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, typeof(*), Tuple{Dagger.BCast{Base.Broadcast.Broadcasted{Dagger.DaggerBroadcastStyle, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, typeof(+), Tuple{Dagger.AllocateArray{Float64, 2}, Dagger.AllocateArray{Float64, 2}}}, Float64, 2}, Int64}}, Float64, 2}(100, 100)
+
```
jpsamaroo, line 86: This would be a good time to point out how the types start getting longer as we start effectively nesting operations.

docs/src/darray.md
```diff
@@ -0,0 +1,135 @@
...
+julia> size(DZ)
+(100, 100)
+
```
jpsamaroo, line 89: Point out that the size is calculated based on set of combined operations, even though the operations haven't yet been computed?

docs/src/darray.md
```diff
@@ -0,0 +1,135 @@
...
+julia> DA = compute(DZ)
+Dagger.DArray{Any, 2, typeof(cat)}(100, 100)
+
```
jpsamaroo, line 92: Point out that now `DA == (DX .+ DX) .* 3`

docs/src/darray.md
```diff
@@ -0,0 +1,135 @@
...
+julia> Dagger.chunks(DA)
+2×2 Matrix{Union{Thunk, Dagger.Chunk}}:
+ Chunk{Matrix{Float64}, DRef, ThreadProc, AnyScope}(Matrix{Float64}, ArrayDomain{2}((1:50, 1:50)), DRef(4, 8, 0x0000000000004e20), ThreadProc(4, 1), AnyScope(), true)  …  Chunk{Matrix{Float64}, DRef, ThreadProc, AnyScope}(Matrix{Float64}, ArrayDomain{2}((1:50, 1:50)), DRef(2, 5, 0x0000000000004e20), ThreadProc(2, 1), AnyScope(), true)
+ Chunk{Matrix{Float64}, DRef, ThreadProc, AnyScope}(Matrix{Float64}, ArrayDomain{2}((1:50, 1:50)), DRef(5, 5, 0x0000000000004e20), ThreadProc(5, 1), AnyScope(), true)     Chunk{Matrix{Float64}, DRef, ThreadProc, AnyScope}(Matrix{Float64}, ArrayDomain{2}((1:50, 1:50)), DRef(3, 3, 0x0000000000004e20), ThreadProc(3, 1), AnyScope(), true)
+
```
jpsamaroo, line 97: Maybe a brief mention that these "chunks" now point to the actual result of the above calculations, split into blocks, but that users don't typically need to worry about these details?

docs/src/darray.md
```diff
@@ -0,0 +1,135 @@
...
+julia> DA *= 2
+Dagger.BCast{Base.Broadcast.Broadcasted{Dagger.DaggerBroadcastStyle, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, typeof(*), Tuple{Dagger.DArray{Any, 2, typeof(cat)}, Int64}}, Any, 2}(100, 100)
+```
+
```
jpsamaroo, line 101: Point out that all DArray operations are functional/immutable, so a new DArray is allocated for the result of every operation?

jpsamaroo (commented):
Great work, thanks for this! I'd like to slightly expand the discussion of the examples to help users understand a little more of what's going on behind the API calls.

docs/src/darray.md
mkitti: Just to make sure. Is "allocated" the right term here? Are we allocating memory on the heap?

jpsamaroo: "created" would be fine, or similar. Whether it's heap allocated or not is mostly irrelevant.

jpsamaroo:
@mkitti sorry for letting this PR go cold! With the recent `DArray` changes, and the docs restructuring that just got merged, I'm going to pick this one up and get it ready to merge.

docs/src/darray.md
jpsamaroo: This behavior will go away by the next release, so I'll skip doing this.

jpsamaroo:
Thanks for this @mkitti !

PR #342, opened 2022-04-13
add_thunk: Optimize domination check

### LINKED ISSUE #341: Execution slow for a simple sequence
I noticed the following code is super slow, I'm using a simple sequence of y(n) = .3*y(n-2) + .7*y(n-1) to test Dagger:

```
using Dagger
fibsum(a, b)=.3*a + .7*b

f0 = Dagger.@spawn rand(10)
f1 = Dagger.@spawn rand(10)
f2 = 0. # result will be held here

@time for i in 1:40
    f2 = Dagger.@spawn fibsum(f0, f1)
    f0 = f1
    f1 = f2
end
```

The code enclosed in @time never returns for me. But if I only compute up to term 20, it does return and produce good results. Anyone else have seen this issue? Am I doing something wrong here?

I'm using Julia 1.6.3 and Dagger v0.14.3,

jpsamaroo:
`dominates` (called in `add_thunk!` -> `register_future!`) previously used a duplicative recursive implementation, which has known issues. This switches `dominates` to a visitor pattern, which results in far better performance on large task trees.

Fixes #341

### DIFF
--- a/src/sch/dynamic.jl
+++ b/src/sch/dynamic.jl
@@ -132,10 +132,21 @@ function _register_future!(ctx, state, task, tid, (future, id)::Tuple{ThunkFutur
         ownthunk = unwrap_weak_checked(state.thunk_dict[tid])
         function dominates(target, t)
             t == target && return true
-            # N.B. Skips expired tasks
-            task_inputs = filter(istask, Dagger.unwrap_weak.(t.inputs))
-            if any(_t->dominates(target, _t), task_inputs)
-                return true
+            seen = Set{Thunk}()
+            to_visit = Thunk[t]
+            while !isempty(to_visit)
+                t = pop!(to_visit)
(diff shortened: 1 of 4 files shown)


jpsamaroo:
@DrChainsaw the "Remove workers" test is being flaky again, so I'm disabling it for now. I think we should sit down and rethink those tests (and potentially Dagger's own add/remove interfaces) to ensure that we can reliably test that behavior, regardless of what decisions the scheduler makes.

DrChainsaw:
> "Remove workers" test is being flaky again, so I'm disabling it for now.

Yeah, I guess it was only a matter of time. I should probably have used better judgement and not tried to push that "fix" :) 

Perhaps just let it be untested until work stealing is implemented as the add/remove functionality anyways is a bit pointless without it?

PR #344, opened 2022-04-19
[DTable] DataFrames-like `select` using DataFrames syntax

krynju:
mega wip


krynju:
Already merged within https://github.com/JuliaParallel/DTables.jl

PR #350, opened 2022-07-02
Remove `poolset` kwarg that doesn't exist anymore

krynju:
also bump MemPool

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -21,7 +21,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 [compat]
 Colors = "0.10, 0.11, 0.12"
 ContextVariablesX = "0.1"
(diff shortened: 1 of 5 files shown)


Project.toml
```diff
@@ -21,7 +21,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 [compat]
 Colors = "0.10, 0.11, 0.12"
 ContextVariablesX = "0.1"
-MemPool = "0.3.5"
+MemPool = "0.4.1"
```
jpsamaroo, line 24:
```suggestion
MemPool = "0.3.5, 0.4"
```

We should be able to support both with this change (the kwarg was dead anyway).

jpsamaroo (approved):

jpsamaroo:
Thanks!

PR #361, opened 2022-11-06
`enable_disk_caching` function for handling MemPool storage setup

krynju:
adds `enable_disk_caching!` which sets up disk caching on every process

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -17,10 +17,10 @@ uuid = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 version = "1.16.0"
 
 [[deps.ChangesOfVariables]]
-deps = ["LinearAlgebra", "Test"]
-git-tree-sha1 = "f84967c4497e0e1955f9a582c232b02847c5f589"
+deps = ["InverseFunctions", "LinearAlgebra", "Test"]
+git-tree-sha1 = "2fba81a302a7be671aefe194f0525ef231104e7f"
 uuid = "9e997f8a-9a97-42d5-a9f1-ce6bfc15e2c0"
-version = "0.1.7"
+version = "0.1.8"
 
 [[deps.ColorTypes]]
 deps = ["FixedPointNumbers", "Random"]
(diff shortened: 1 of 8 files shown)


krynju:
Yeah, I was experimenting with this and just put this up
I initially added it to MemPool but then i thought this would be better placed in Dagger

Now I think maybe we should have a single process `enable_caching` function in MemPool and then from Dagger level have a wrapper to enable it on all workers?
In DTables i would just add a helper function to call the Dagger implementation 

What do you think?

jpsamaroo:
That all sounds perfect to me!

src/utils/caching.jl
```diff
@@ -0,0 +1,104 @@
...
+        true
+    end
+end
+
+"""
+    inspect_global_devices(processes::Vector{Int}=procs())
+
+Helper function that returns the MemPool `GLOBAL_DEVICE` from each process.
+"""
+function inspect_global_devices(processes::Vector{Int}=procs())
```
jpsamaroo, line 96: This isn't safe to do, as storage devices aren't intended to be serializable (it can easily cause serialization of all the data for `DRef`s contained in the LRU). Do we need this for anything specific?

src/utils/caching.jl
```diff
@@ -0,0 +1,104 @@
...
+        if r !== nothing
+            @error("Error setting up disk caching on process id = $id", ex = r)
+        end
+    end
+
+    return if any_error
+        @error("Disk cache setup failed")
+        false
+    else
+        @info("Disk cache setup successful")
```
jpsamaroo, line 86:
```suggestion
```

I prefer to avoid printing unless there's something exceptional happening (which then usually warrant a warning or error instead).

Manifest.toml
jpsamaroo, line 1: I'm going to need to put back the Manifest, since it's (annoyingly) required to get CI working.

jpsamaroo (approved):
Awesome!

src/utils/caching.jl
```diff
@@ -0,0 +1,69 @@
+function enable_disk_caching!(
+    ram_percentage_limit::Int=30, disk_limit_gb::Int=16, processes::Vector{Int}=procs()
```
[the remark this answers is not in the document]
jpsamaroo: We can skip this for now, and if people complain, we can add another kwarg that's in terms of `0.0:1.0`.

Manifest.toml
krynju: oh that's weird, we should fix that separately then

src/utils/caching.jl
krynju: Not really, I think I only used it for checking the status quickly

krynju: we can remove it

jpsamaroo:
Thanks!

PR #366, opened 2022-11-13
Add DTables.jl CI run

krynju:
DTables.jl in threaded mode is a very good stability/correctness test for Dagger, which we usually run manually anyway, so let's add it to CI


jpsamaroo:
Thanks, this will be super useful to have!

PR #370, opened 2023-02-03
Fix incorrect assertion in schedule!

jpsamaroo:
Previously, we had an assertion in `schedule!` that a task being scheduled could not already have a cache result, as that seems counterintuitive. However, when an upstream task has thrown an error, downstream tasks are all eagerly tainted by `set_failed!`, which sets their cache value to the `ThunkFailedException` object.

This PR loosens that assertion to ignore errored tasks, and refines the non-erroring case to throw a `SchedulingException` to the task instead.


jpsamaroo:
Also, this improves `ThunkFailedException` printing, which was previously far too verbose for wide or deep DAGs.

Issue #379, opened 2023-03-03
TimespanLogging: Remove worker ID nesting


jpsamaroo:
It's annoying that `get_logs!` of a `MultiEventLog` returns an extra level of nesting by worker ID; logging is really process-local, so we should remove that nesting and let the user pull logs from multiple workers if needed.

PR #387, opened 2023-03-17
TimespanLogging: Remove worker Id Nesting

Madhupatel08:
Removed the extra level of nesting by worker ID (#379)
Output when only_local is true: logs are fetched only from the current process. 
Logs can be directly fetched from consumer_logs since there is only one worker.  
image

Output when only only_local is false: logs are fetched from all the workers
image

### DIFF
--- a/lib/TimespanLogging/src/core.jl
+++ b/lib/TimespanLogging/src/core.jl
@@ -258,7 +258,7 @@ function get_logs!(ml::MultiEventLog; only_local=true)
             end
         end
         return logs
-    end       
+    end
 end
 
 
--- a/lib/TimespanLogging/src/core.jl
+++ b/lib/TimespanLogging/src/core.jl
@@ -235,21 +235,33 @@ end
 
 function get_logs!(ml::MultiEventLog; only_local=true)
     if only_local
-        # Only return logs from current process
-        mls = get_state(ml)
-        return mls.consumer_logs
+        # Only return logs from the current process
+        logs = Dict{Int, Dict{Symbol,Vector}}()
+        pid = myid()
+        logs[pid] = remotecall_fetch(pid, ml) do ml 
+            mls = get_state(ml)
+            lock(event_log_lock) do
+                sublogs = Dict{Symbol,Vector}()
+                for name in keys(mls.consumers)
+                    sublogs[name] = copy(mls.consumer_logs[name])
+                    mls.consumer_logs[name] = []
+                end
+                sublogs
+            end
+        end
+        return logs
     else
         # Return logs from all processes
-        logs = Dict{Int,Dict{Symbol,Vector}}()
+        logs = Dict{Int, Dict{Symbol,Vector}}()
         wkrs = procs()
-        @sync for p in wkrs 
+        @sync for p in wkrs
             @async begin
-                logs[p] = remotecall_fetch(p, ml) do ml
+                logs[p] = remotecall_fetch(p, ml) do ml 
                     mls = get_state(ml)
                     lock(event_log_lock) do
                         sublogs = Dict{Symbol,Vector}()
                         for name in keys(mls.consumers)
-                            sublogs[name] = mls.consumer_logs[name]
+                            sublogs[name] = copy(mls.consumer_logs[name])
                             mls.consumer_logs[name] = []
                         end
                         sublogs
@@ -478,4 +490,3 @@ function summarize_events(time_spent, gc_diff, profiler_samples)
 end
 
 summarize_events(xs) = summarize_events(aggregate_events(xs)...)
-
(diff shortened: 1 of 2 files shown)


lib/TimespanLogging/src/core.jl
```diff
@@ -233,28 +233,35 @@ function write_event(ml::MultiEventLog, event::Event)
...
+                        sublogs
                     end
-                    sublogs
                 end
             end
         end
-    end
-    return logs
+        return logs
+    end
```
jpsamaroo, line 261:
```suggestion
    end
```

Extra whitespace

lib/TimespanLogging/src/core.jl
```diff
@@ -233,28 +233,35 @@ function write_event(ml::MultiEventLog, event::Event)
...
-                lock(event_log_lock) do
-                    sublogs = Dict{Symbol,Vector}()
-                    for name in keys(mls.consumers)
-                        sublogs[name] = mls.consumer_logs[name]
-                        mls.consumer_logs[name] = []
+function get_logs!(ml::MultiEventLog; only_local=true)
+    if only_local
+        # Only return logs from current process
+        mls = get_state(ml)
+        return mls.consumer_logs
```
jpsamaroo, line 240: We need to make a copy of these logs and then empty each vector contained in `mls.consumer_logs`, as the API of `get_logs!` is to remove the old logs from `mls.consumer_logs` so that future calls only include new logs.

jpsamaroo (commented):
Looking good so far! Just two changes needed here, and then you'll also need to update the tests to ensure we test both `only_local=true` and `only_local=false`.

lib/TimespanLogging/src/core.jl
Madhupatel08: okay will fix it

lib/TimespanLogging/src/core.jl
```diff
@@ -233,28 +233,47 @@ function write_event(ml::MultiEventLog, event::Event)
...
-                    sublogs = Dict{Symbol,Vector}()
-                    for name in keys(mls.consumers)
-                        sublogs[name] = mls.consumer_logs[name]
-                        mls.consumer_logs[name] = []
+function get_logs!(ml::MultiEventLog; only_local=true)
+    if only_local
+        # Only return logs from the current process
+        logs = Dict{Int, Dict{Symbol,Vector}}()
+        pid = myid()
+        logs[pid] = remotecall_fetch(pid, ml) do ml
```
jpsamaroo, line 241: Why are you `remotecall_fetch`-ing here? The worker calling `get_logs!` with `only_local=true` is by definition the same worker that you need to fetch logs from, so you don't need that level of indirection.

lib/TimespanLogging/src/core.jl
```diff
@@ -233,28 +233,47 @@ function write_event(ml::MultiEventLog, event::Event)
...
+            lock(event_log_lock) do
+                sublogs = Dict{Symbol,Vector}()
+                for name in keys(mls.consumers)
+                    sublogs[name] = copy(mls.consumer_logs[name])
+                    mls.consumer_logs[name] = []
+                end
+                sublogs
+            end
+        end
+        return logs
```
jpsamaroo, line 252: This is wrong, it's now returning logs with an extra layer of nesting.

test/logging.jl
```diff
@@ -60,74 +61,85 @@ import TimespanLogging: Timespan, Event, Events, LocalEventLog, MultiEventLog
...
+            ml[:wsat] = Dagger.Events.WorkerSaturation()
+            ml[:loadavg] = Events.CPULoadAverages()
+            ml[:bytes] = Dagger.Events.BytesAllocd()
+            ml[:mem] = Events.MemoryFree()
+            ml[:esat] = Events.EventSaturation()
+            ml[:psat] = Dagger.Events.ProcessorSaturation()
+            ctx.log_sink = ml
+    
+            A = rand(Blocks(4, 4), 16, 16)
+            collect(ctx, A*A)
```
jpsamaroo, line 79: These lines need to go inside the loop, to ensure we're generating logs for each of `only_local=true` and `only_local=false`.

test/logging.jl
```diff
@@ -60,74 +61,85 @@ import TimespanLogging: Timespan, Event, Events, LocalEventLog, MultiEventLog
...
-            end
-        end
-        @test had_psat_proc > 0
-
-        logs = TimespanLogging.get_logs!(ml)
-        for w in keys(logs)
-            for c in keys(logs[w])
-                @test isempty(logs[w][c])
-            end
+        @testset "Get logs for only_local = true and only_local = false" begin
```
jpsamaroo, line 64: This extra level of testset nesting isn't necessary.

jpsamaroo (changes requested):

PR #390, opened 2023-03-27
Updating Task cost estimation

Madhupatel08:
Changed the costs dictionary to calculate the cost for each worker to include the cost of moving non-Chunk inputs from the scheduler to the worker. We sum the data size of all non-Chunk inputs and calculate the cost of moving that data from the scheduler to the worker using the transfer rate and then add that cost to the state.scheduler_time_pressure to get the total cost for the scheduler.

### DIFF
--- a/src/sch/util.jl
+++ b/src/sch/util.jl
@@ -399,8 +399,16 @@ function estimate_task_costs(state, procs, task, inputs)
     transfer_costs = Dict(proc=>impute_sum([affinity(chunk)[2] for chunk in filter(c->get_parent(processor(c))!=get_parent(proc), chunks)]) for proc in procs)
 
     # Estimate total cost to move data and get task running after currently-scheduled tasks
-    costs = Dict(proc=>state.worker_time_pressure[get_parent(proc).pid][proc]+(tx_cost/tx_rate) for (proc, tx_cost) in transfer_costs)
-
+    costs = Dict(proc-> begin
+                        if proc == state.scheduler
+                            # For non-Chunk, model cost from scheduler to worker
+                            data_size = sum(affinity(input)[2] for input in filter(x -> !(x isa Chunk), inputs))
+                            scheduler_to_worker_cost = (data_size / tx_rate) + state.scheduler_time_pressure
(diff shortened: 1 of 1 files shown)


src/sch/util.jl
```diff
@@ -399,8 +399,16 @@ function estimate_task_costs(state, procs, task, inputs)
     transfer_costs = Dict(proc=>impute_sum([affinity(chunk)[2] for chunk in filter(c->get_parent(processor(c))!=get_parent(proc), chunks)]) for proc in procs)
 
     # Estimate total cost to move data and get task running after currently-scheduled tasks
-    costs = Dict(proc=>state.worker_time_pressure[get_parent(proc).pid][proc]+(tx_cost/tx_rate) for (proc, tx_cost) in transfer_costs)
-
+    costs = Dict(proc-> begin
+                        if proc == state.scheduler
```
jpsamaroo, line 403: What is `state.scheduler`?

src/sch/util.jl
```diff
@@ -399,8 +399,16 @@ function estimate_task_costs(state, procs, task, inputs)
...
+                            # For non-Chunk, model cost from scheduler to worker
+                            data_size = sum(affinity(input)[2] for input in filter(x -> !(x isa Chunk), inputs))
```
jpsamaroo, line 405:
```suggestion
                            data_size = sum(MemPool.approx_size(input) for input in filter(x -> !(x isa Chunk), inputs))
```

`affinity` only works on special objects like `Chunk`, `Thunk`, etc. `MemPool.approx_size` works on any kind of object.

src/sch/util.jl
```diff
@@ -399,8 +399,16 @@ function estimate_task_costs(state, procs, task, inputs)
...
+                            scheduler_to_worker_cost = (data_size / tx_rate) + state.scheduler_time_pressure
```
jpsamaroo, line 406: There is no scheduler-specific time pressure, you can access that with just `state.worker_time_pressure[1]` (as the scheduler runs on worker 1). Therefore, you don't need to deal with that because it's already factored in on the next line.

jpsamaroo (changes requested):

jpsamaroo:
Closing as this is stale

PR #405, opened 2023-07-19
DArray: MPI interface

fda-tome:
Implementing reductions using MPI collective calls an interfacing the DArray to use MPI style instead of PGAS style (Partitioned global address space)


jpsamaroo:
Superseded by #422

PR #406, opened 2023-07-19
DArray: New Blocks and correct returns

fda-tome:
New AbstractBlocks to enhance further development and functions that are expected to return a DArray now do it correctly


jpsamaroo:
Implemented in #408

PR #408, opened 2023-07-20
DArray: Operations always return DArrays, and add local partition support for MPI

jpsamaroo:
Rebase and tweaks of #406, and prep for DaggerMPI support


jpsamaroo:
Great work, thanks you @fda-tome !

PR #418, opened 2023-08-08
Update Dagger README.md docs with a use case sceanario

pszufe:
@jpsamaroo here is a use case scenario for dagger usage. It demonstrates a simplified computational scenario, yet stylized after real world settings. The scenario immediately demonstrates the advantage of using Dagger over the `Threads.@threads` macro.


jpsamaroo:
Thanks @pszufe !

PR #420, opened 2023-08-09
Use package extensions

oschulz:


oschulz:
@jpsamaroo is the a chance that Dagger will support Julia v1.6 (again)? If not, should we just require Julia v1.9 instead of v1.7 and get rid of Requires completely?

jpsamaroo:
As discussed on Slack, we want to support 1.7+ for the time being.

oschulz:
Ok, I'll keep Requires in then and try to fix the CI failures.

PR #422, opened 2023-08-15
MPI DArray: finished distribute, reduce and allocations

fda-tome:
This PR wraps up the base version of the DArray for MPI, including randomized allocations, ones, zeros, distribute without specified partition dimensions, and all the variations of reduce (sum, prod, reduce with a root, and without a root), including reduce dims and all-collect.


jpsamaroo:
Implemented in #720

PR #425, opened 2023-08-17
Update path splitting rule in Dagger.File / Dagger.tofile

svilupp:
This PR proposes a fix to a potential bug discovered in https://github.com/JuliaParallel/DTables.jl/issues/35

**Behaviour:** When using relative paths (eg, `Dagger.tofile(data, "data_processed/obj_$i.arrow", ...)`, the directories would be ignored and files would be saved directly in pwd().

**Reason:** Path splitting was done with `isabspath` (see [link](https://github.com/JuliaParallel/Dagger.jl/blob/04e2466a92927faecb9d1c27cf50f5b93bc8df3b/src/file-io.jl#L33), but that doesn't work for relative paths.

Example:
```julia
path = "data_output/out.arrow"
isabspath(path) # false! not helpful if we want to know if we need to use pwd() to set directory!
```

Proposed change:
```julia
dir, file = dirname(path), basename(path)
dir = isempty(dir) ? pwd() : dir
```

I've also added a test case for the relative paths to prevent future regressions. Current testing used `tempdir()`, which provides an absolute path; hence, the bug wasn't observed.

### DIFF
--- a/src/file-io.jl
+++ b/src/file-io.jl
@@ -30,11 +30,9 @@ function File(path::AbstractString;
         # FIXME: Once MemPool better propagates errors, this check won't be necessary
         throw(ArgumentError("`Dagger.File` expected file to exist at \"$path\""))
     end
-    if isabspath(path)
-        dir, file = dirname(path), basename(path)
-    else
-        dir, file = pwd(), basename(path)
-    end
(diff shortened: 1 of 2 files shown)


jpsamaroo:
Thanks for finding and fixing this!

PR #451, opened 2023-11-03
DaggerWebDash: Use gethostname and getipaddr

jpsamaroo:
We were using `MemPool.host` (which no longer exists), when really we should just call `gethostname()` or `getipaddr()`.


mofeing:
`gethostname()` returns a `String` but a `Sockets.IpAddr` is required.

Replacing it with `Sockets.getipaddr()` solves the problem.

jpsamaroo:
I had to widen the internal IP/host cache type to accommodate `String`. I also decided to vendor d3.js, since it's annoying to need an internet connection just to use a purely-local visualizer.

PR #465, opened 2024-01-23
Sch: Temporarily disable chunk cache

jpsamaroo:
The scheduler chunk cache has been around for a while, but unfortunately was never really correct in the face of mutating data. If data is mutated by any Dagger task, Dagger's chunk cache may still serve the stale, non-mutated data copy. This issue has come up in a number of places, and so it needs to be resolved now before causing further issues. A proper fix is still in the works (as part of a larger set of changes), but in the meantime, this hotfix will have to do.

### DIFF
--- a/src/sch/Sch.jl
+++ b/src/sch/Sch.jl
@@ -1508,6 +1508,7 @@ function do_task(to_proc, task_desc)
     fetch_tasks = map(Iterators.zip(_data,_ids)) do (x, id)
         @async begin
             timespan_start(ctx, :move, (;thunk_id, id), (;f, id, data=x))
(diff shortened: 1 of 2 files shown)


jpsamaroo:
Aside: according to Dagger's current semantics (that tasks are functional in nature, and so arguments should not be mutated), this is technically correct behavior, but it is also confusing and hard to disable when you do need mutation. It's also correct to *not* cache chunk values, so it's valid to remove it. Future changes will enable mutation of arguments with some simple argument annotations, so this work is really preparing for that, and avoiding a quiet, hard-to-detect footgun until that support lands.

PR #469, opened 2024-02-03
DArray: Add Cholesky and view implementations

jpsamaroo:
Tiled Cholesky implementation by @rabab53


test/linalg.jl
```diff
@@ -0,0 +1,34 @@
+@testset "Linear Algebra" begin
+    @testset "Cholesky: $T" for T in (Float32, Float64, ComplexF64)
```
Rabab53, line 2: Even ComplexF32 should work too

src/array/cholesky.jl
```diff
@@ -0,0 +1,82 @@
...
+        for j in 1:j_pre_diag
+            push!(to_check, Dagger.@spawn is_cross_symmetric(Ac[i, j], Ac[j, i]))
+        end
+    end
+
+    return all(fetch, to_check)
+end
+
+LinearAlgebra.cholcopy(A::DArray{T,2}) where T = copy(A)
+function LinearAlgebra.cholesky!(A::DArray{T,2}, ::NoPivot; check::Bool=true) where T
```
Rabab53, line 36: I think we have to be explicit when using `check` word, because it could refer to correctness of factorization or of the matrix specific properties

src/array/cholesky.jl
```diff
@@ -0,0 +1,82 @@
...
+        j_pre_diag = i - 1
+        for j in 1:j_pre_diag
+            push!(to_check, Dagger.@spawn is_cross_symmetric(Ac[i, j], Ac[j, i]))
+        end
+    end
+
+    return all(fetch, to_check)
+end
+
+LinearAlgebra.cholcopy(A::DArray{T,2}) where T = copy(A)
```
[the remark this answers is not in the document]
jpsamaroo: Seems like it's a required part of the interface, and our `getindex` fallback is broken and inefficient.

src/array/cholesky.jl
jpsamaroo: What would be a better word to use than `check`? That's what the `cholesky!(::Array)` implementation in LinearAlgebra uses.

src/array/cholesky.jl
```diff
@@ -0,0 +1,82 @@
...
+                for n in range(k+1, nt)
+                    Dagger.@spawn BLAS.syrk!(uplo, 'N', mzone, In(Ac[n, k]), zone, InOut(Ac[n, n]))
+                    for m in range(n+1, mt)
+                        Dagger.@spawn BLAS.gemm!('N', 'T', mzone, In(Ac[m, k]), In(Ac[n, k]), zone, InOut(Ac[m, n]))
+                    end
+                end
+            end
+        end
+
+        return Cholesky(A, 'L', 0)
```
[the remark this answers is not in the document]
jpsamaroo: Resolved with checked `potrf!` implementation

src/array/cholesky.jl
jpsamaroo: No longer used in the latest commits

Issue #444, opened 2023-10-25
How to spawn workers over ssh?


KronosTheLate:
Can I use Dagger to spawn processes on another computer connected over ssh? 

`Distributed.addprocs` has a method that "Add worker processes on remote machines via SSH.". Can spawn workers with this, schedule tasks on them with Dagger, and expect it to work?

A few sentences on ssh and computations on other computers in the docs would be lovely ^_^

jpsamaroo:
Yes, you would use the standard `addprocs` from Distributed to do this. Feel free to add a sentence or two to the docs to point this out (maybe in the Quickstart section?) if you have time :slightly_smiling_face:

PR #483, opened 2024-03-13
DOC: Spawning workers over SSH

Schefflera-Arboricola:
closes https://github.com/JuliaParallel/Dagger.jl/issues/444

Let me know if this is the right place to put it.
And I didn't test it as I don't have a remote machine. I referred these docs: https://docs.julialang.org/en/v1/stdlib/Distributed/

Thank you :)


jpsamaroo (changes requested):
Dagger doesn't provide this functionality, this is provided by Distributed.jl (which documents how to spawn workers). This would be `Distributed.addprocs` (or just `addprocs` if the user has done `using Distributed`).

PR #484, opened 2024-03-13
Reference Dagger.EAGER_THUNK_STREAMS explicitly

JamesWrigley:
A lot of tests failed when I was running them locally because that variable is defined in the `Dagger` module rather than `Dagger.Sch`. Though a lot still fail for me with this error:
```julia
  Got exception outside of a @test                                                                                                                                                                                                             
  On worker 2:                                                                                                                                                                                                                                 
  Tuple field type cannot be Union{}
```

I think we're running into https://github.com/JuliaLang/julia/issues/52385 :grimacing: I'll try to look into that in a few days.

(P.S.: I configured the PR to merge into #463 instead of `master`)

### DIFF
--- a/src/sch/eager.jl
+++ b/src/sch/eager.jl
@@ -117,7 +117,7 @@ function eager_cleanup(state, uid)
         delete!(state.thunk_dict, tid)
     end
     remotecall_wait(1, uid) do uid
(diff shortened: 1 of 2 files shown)


jpsamaroo:
Thanks, this avoids the heartburn I feel when using `Base._return_type` :smile:

PR #486, opened 2024-03-17
Special-case StreamingFunction in EagerThunkMetadata() constructor

JamesWrigley:
This was using the `chunktypes()` API, which I don't think is correct in general. Now all the tests pass for me.


jpsamaroo:
Thanks!

Issue #482, opened 2024-03-11
Updating to Documenter.jl v1


JamesWrigley:
I attempted to update the docs to build with Documenter.jl v1, but it seems that there are dependency conflicts. I narrowed it down to:
- The docs project depending on `DaggerWebDash`
  - Which depends on [ProfileSVG.jl](https://github.com/kimikage/ProfileSVG.jl)
    - Which has a [compat bound](https://github.com/kimikage/ProfileSVG.jl/blob/master/Project.toml#L13) to [FlameGraphs.jl](https://github.com/timholy/FlameGraphs.jl) v0.2 (an old version)
      - Which has a [compat bound](https://github.com/timholy/FlameGraphs.jl/blob/v0.2.10/Project.toml#L16) to an old version of AbstractTrees.jl
        - Which is incompatible with Documenter.jl v1: https://github.com/JuliaDocs/Documenter.jl/blob/master/Project.toml#L32
        
This PR would fix it: https://github.com/kimikage/ProfileSVG.jl/pull/70
But I'm not sure if that'll get merged anytime soon. Anyway it's not urgent, I just wanted to document the chain of pain :grin:

jpsamaroo:
Thanks for looking into this! I'd be ok with removing ProfileSVG support from DaggerWebDash for now - it's not really used, since profiling support in Dagger isn't great, because Profile doesn't support continuous profiling (yet another chain of pain :laughing: ).

asinghvi17:
If it's a web viewer, you could try switching to ProfileCanvas.jl instead?  Not sure if it needs a live server, though.

Alternatively, you can always just add that branch in CI explicitly!

jpsamaroo:
The live rendering support is enabled by running ProfileSVG multiple times, so ProfileCanvas.jl could work! Thanks for letting me know about it :smile:

JamesWrigley:
Fixed in #487.

PR #487, opened 2024-03-18
Stop tracking the manifest files

JamesWrigley:
I don't think these really need to be tracked, and they cause conflicts whenever the manifest is changed and you switch branch or something.


jpsamaroo:
This generally LGTM - do you know why the docs build is failing? Maybe the default settings changed upstream? We need to be able to skip some of these missing doc checks until we get that fully fixed up.

JamesWrigley:
Ahh yes, that'll be because #482 is fixed now that ProfileSVG.jl has been updated, so now the docs will be using Documenter.jl v1 which has stricter defaults. I'll update the docs :)

JamesWrigley:
Fixed in 518a4b0 and 85df3cb, so this should fix #482 too now.

JamesWrigley:
A bunch of other jobs are still broken, but I don't that's caused by the manifest being deleted, and it should be working in #463, sooooo I don't think I'll bother fixing that :see_no_evil:

jpsamaroo:
Thanks!

Yes those jobs were previously failing, but I haven't had the chance to look into them.

PR #492, opened 2024-03-26
DArray: Implement efficient in-place matrix-matrix multiply

Rabab53:
This PR implements:
- Optimized `mul!` and `*` using datadeps
- Optimized `transpose`/`adjoint` using datadeps
- `copyto!` using datadeps
- Optimized `getindex` with optional task-local caching
- `setindex!` (to be optimized later)
- A convenient test runner script (used like `julia test/runtests.jl --test array/linalg/matmul --test array/copies`)


jpsamaroo:
Thanks @Rabab53!

PR #495, opened 2024-03-30
Doc/add Quickstart datadeps documentation

johnmaxrin:
Add Quickstart  documentation for Datadeps in Dagger.jl

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -315,3 +315,52 @@ To get back an `Array` from a `DArray`, just call `collect`:
 DA = rand(Blocks(32, 32), 256, 128)
 collect(DA) # returns a `Matrix{Float64}`
 ```
+
+## Quickstart: Datadeps
+
+Datadeps is a feature in Dagger.jl that facilitates parallelism control within designated regions, allowing tasks to write to their arguments while ensuring dependencies are respected.
+
+### Syntax
+The Dagger.spawn_datadeps() function is used to create a "datadeps region" where tasks are executed with parallelism controlled by specified dependencies.
+
+```julia
+Dagger.spawn_datadeps() do
+    Dagger.@spawn func!(Out(var_name_x), In(var_name_y))
+end
+```
+### Argument Annotation
+
+- In(X): Indicates that the variable X is an input to the task.
+- Out(X): Indicates that the variable X is an output of the task.
+- InOut: Indicates that the variable is both an input and an output.
+
+### Example with Datadeps
+
+```julia
+X = [4,5,6,7,1,2,3,9,8]
+C = zeros(10)
+
+Dagger.spawn_datadeps() do
+    Dagger.@spawn sort!(InOut(X))
+    Dagger.@spawn copyto!(Out(C), In(X))
+end
+
+# C = [1,2,3,4,5,6,7,8,9]
+```
+
+In this example, the sort! function operates on array X, while the copyto! task operates on array C. By specifying dependencies using argument annotations, the tasks are executed in a controlled parallel manner, resulting in a sorted C array.
+
+### Example without Datadeps
+
+```julia
+X = [4,5,6,7,1,2,3,9,8]
+C = zeros(10)
+
+Dagger.@spawn sort!(InOut(X))
+Dagger.@spawn copyto!(Out(C), In(X))
+
+# C = [4,5,6,7,1,2,3,9,8]
+```
+
+In contrast to the previous example, here, the tasks are executed without Datadeps and argument annotations. As a result, there is a possibility of the copyto! task being executed before the sort! task, leading to unexpected behavior in the output array C.
+
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -356,12 +356,14 @@ In this example, the sort! function operates on array X, while the copyto! task 
 ```julia
 X = [4,5,6,7,1,2,3,9,8]
 C = zeros(10)
+Dagger.spawn_datadeps() do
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -315,3 +315,55 @@ To get back an `Array` from a `DArray`, just call `collect`:
...
+```julia
+Dagger.spawn_datadeps() do
+    Dagger.@spawn func!(Out(var_name_x), In(var_name_y))
+end
+```
+### Argument Annotation
+
+- In(X): Indicates that the variable X is an input to the task.
+- Out(X): Indicates that the variable X is an output of the task.
+- InOut: Indicates that the variable is both an input and an output.
```
jpsamaroo, line 336:
```suggestion
- `In(X)`: Indicates that the variable `X` is only read by the task (an "input").
- `Out(X)`: Indicates that the variable `X` is only written to by the task (an "output").
- `InOut(X)`: Indicates that the variable `X` is both read from and written to by the task (an "input" and "output" simultaneously).
```

To indicate what "input" and "output" really signify as far as task behavior is concerned.

docs/src/index.md
```diff
@@ -315,3 +315,55 @@ To get back an `Array` from a `DArray`, just call `collect`:
...
+
+### Example without Datadeps
+
+```julia
+X = [4,5,6,7,1,2,3,9,8]
+C = zeros(10)
+Dagger.spawn_datadeps() do
+    Dagger.@spawn sort!(X)
+    Dagger.@spawn copyto!(C, X)
+end
```
jpsamaroo, line 362:
```suggestion
Dagger.@spawn sort!(X)
Dagger.@spawn copyto!(C, X)
```

Technically it's functionally identical, but I think it's a bit confusing if we say "without Datadeps", but then still have the `spawn_datadeps` region.

docs/src/index.md
```diff
@@ -315,3 +315,55 @@ To get back an `Array` from a `DArray`, just call `collect`:
...
+
+Dagger.spawn_datadeps() do
+    Dagger.@spawn sort!(InOut(X))
+    Dagger.@spawn copyto!(Out(C), In(X))
+end
+
+# C = [1,2,3,4,5,6,7,8,9]
+```
+
+In this example, the sort! function operates on array X, while the copyto! task operates on array C. By specifying dependencies using argument annotations, the tasks are executed in a controlled parallel manner, resulting in a sorted C array.
```
jpsamaroo, line 352:
```suggestion
In this example, the `sort!` function operates on array `X`, while the `copyto!` task reads from array `X` and reads from array `C`. By specifying dependencies using argument annotations, the tasks are executed in a controlled parallel manner, resulting in a sorted `C` array.
```

docs/src/index.md
jpsamaroo: Alternatively, you could leave `spawn_datadeps` here, but change the header to "Example with Datadeps (without annotations)" or something like that.

docs/src/index.md
```diff
@@ -315,3 +315,55 @@ To get back an `Array` from a `DArray`, just call `collect`:
...
+Dagger.spawn_datadeps() do
+    Dagger.@spawn sort!(X)
+    Dagger.@spawn copyto!(C, X)
+end
+
+
+# C = [4,5,6,7,1,2,3,9,8]
+```
+
+In contrast to the previous example, here, the tasks are executed without argument annotations. As a result, there is a possibility of the copyto! task being executed before the sort! task, leading to unexpected behavior in the output array C.
```
jpsamaroo, line 368:
```suggestion
In contrast to the previous example, here, the tasks are executed without argument annotations. As a result, there is a possibility of the `copyto!` task being executed before the sort! task, leading to unexpected results in the output array `C`.
```

jpsamaroo (commented):
Great stuff! I like the example with `sort!` and `copyto!`, it's simple yet motivating :smile:

johnmaxrin (commented):
Done.

docs/src/index.md
```diff
@@ -315,3 +315,55 @@ To get back an `Array` from a `DArray`, just call `collect`:
...
+X = [4,5,6,7,1,2,3,9,8]
+C = zeros(10)
+Dagger.@spawn sort!(X)
+Dagger.@spawn copyto!(C, X)
+
+
+# C = [4,5,6,7,1,2,3,9,8]
+```
+
+In contrast to the previous example, here, the tasks are executed without argument annotations. As a result, there is a possibility of the `copyto!` task being executed before the sort! task, leading to unexpected results in the output array `C`.
```
jpsamaroo, line 368:
```suggestion
In contrast to the previous example, here, the tasks are executed without argument annotations. As a result, there is a possibility of the `copyto!` task being executed before the `sort!` task, leading to unexpected results in the output array `C`.
```

docs/src/index.md
```diff
@@ -315,3 +315,55 @@ To get back an `Array` from a `DArray`, just call `collect`:
 DA = rand(Blocks(32, 32), 256, 128)
 collect(DA) # returns a `Matrix{Float64}`
 ```
+
+## Quickstart: Datadeps
+
+Datadeps is a feature in Dagger.jl that facilitates parallelism control within designated regions, allowing tasks to write to their arguments while ensuring dependencies are respected.
+For more details: [Datadeps](@ref)
```
jpsamaroo, line 322:
```suggestion
For more details: [Datadeps (Data Dependencies)](@ref)
```

jpsamaroo:
Thanks @johnmaxrin !

PR #515, opened 2024-05-16
Add option to the tests for setting the number of workers

JamesWrigley:
This is useful when testing if you don't need extra workers and don't want to wait for them to be started.


jpsamaroo:
Awesome, I'd been meaning to add this. Thanks!

PR #516, opened 2024-05-16
Test fixes

JamesWrigley:
This fixes some issues we're seeing in @davidizzle's tests:
- Sometimes the `test_finishes()` test harness will attempt to cancel a task after the scheduler has already exited, causing exceptions.
- The tests would hang at a certain point if debug logging was turned off.


JamesWrigley:
Discussed this on Zoom with @jpsamaroo, and we agreed it's ok to merge.

PR #518, opened 2024-05-22
Rename EagerThunk to DTask

jpsamaroo:
`EagerThunk` is a terrible name - let's change this to something a bit more easily explainable. Also adds a better docstring for `Dagger.@spawn` to help explain how it differs from Julia's `Task`.

Also deprecates `delayed` - this API is no longer recommended for various reasons, and will be removed as a user-facing API in the future.


src/sch/eager.jl
```diff
@@ -24,7 +24,7 @@ function init_eager()
         sopts = SchedulerOptions(;allow_errors=true)
         opts = Dagger.Options((;scope=Dagger.ExactScope(Dagger.ThreadProc(1, 1)),
                                 occupancy=Dict(Dagger.ThreadProc=>0)))
-        Dagger.compute(ctx, Dagger.delayed(eager_thunk, opts)();
+        Dagger.compute(ctx, Dagger.delayed_internal(eager_thunk, opts)();
```
JamesWrigley, line 27: Nitpicking, what do you think about naming it `_delayed()` instead? I know this literally has `internal` in the name but the leading underscore is also used in Python and `Base` so it might be a bit more recognizable.

JamesWrigley (commented):
Big fan of both of these changes :) I think we can go ahead and bump the version number already since this will be breaking. Another thing I noticed is that a lot of the mentioned symbols in the docstrings could be converted into linked references with e.g. ``[`Dagger.@spawn()`](@ref)``, but I can fix those later.

JamesWrigley:
The array allocation test failure on 1.10 is a little concerning :thinking: Is this the first time it happened or is that just a flaky test?

PR #519, opened 2024-05-24
Add support for limiting the evaluations of a streaming DAG

JamesWrigley:
This is useful for testing and benchmarking.

BTW I added the `stream_max_evals` option to `SchedulerOptions` and `ThunkOptions`, but I see that the other streaming options are sort of used adhoc and filtered out before being passed upwards. Is that just for quick prototyping? Or should the streaming options go somewhere else?

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,7 +5,8 @@
     sandbox_capable: "true"
     os: linux
     arch: x86_64
-  command: "julia --project -e 'using Pkg; Pkg.develop(;path=\"lib/TimespanLogging\")'"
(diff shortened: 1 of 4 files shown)


JamesWrigley:
Also attempted to improve CI by installing the migration-helper branch of MemPool.jl in 4ff2fa1.

JamesWrigley:
I figured out why my filtering was failing, I was doing it in the wrong place :stuck_out_tongue: There was some filtering in `initialize_streaming!()` that I think doesn't actually do anything because those options are in `spec.options.propagate`, not `spec.options`. I removed that filter and added it in the proper place in `finalize_streaming!()` in f634923.

(P.S. if you approve, I'll rebase this before merging to squash the fixup commits)

jpsamaroo (approved):
Looks great, thank you!

JamesWrigley:
The CI seems to be a little overloaded so I'm gonna roguishly go ahead and merge this :pirate_flag:

PR #526, opened 2024-06-01
tests: Fix assumption that rand won't return 0

jpsamaroo:
I'm not sure how this can reliably happen...

### DIFF
--- a/test/array/allocation.jl
+++ b/test/array/allocation.jl
@@ -40,7 +40,8 @@ end
             @test AX == collect(X)
             @test AX != collect(rand(dist, T, dims...))
             if T <: AbstractFloat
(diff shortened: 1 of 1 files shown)


test/array/allocation.jl
```diff
@@ -40,7 +40,8 @@ end
             @test AX == collect(X)
             @test AX != collect(rand(dist, T, dims...))
             if T <: AbstractFloat
-                @test all(AX .> 0)
+                # FIXME: Not ideal, but I guess sometimes we can get 0?
+                @test sum(!.(AX .> 0)) < 10
```
jpsamaroo, line 44:
```suggestion
                @test sum(.!(AX .> 0)) < 10
```

Issue #512, opened 2024-05-14
Using Dagger in an epidemiological analysis: visualising an implied DAG and other examples of usage


SamuelBrand1:
Hi everyone,

I'm implementing an analysis batch for an epidemiological modelling package https://github.com/CDCgov/Rt-without-renewal/tree/main/pipeline .

We're wondering if there is a functionality for graph plotting a DAG after it has been implied (e.g. by using `Dagger.@spawn` to generate a number of `Thunk`s)? I've noted `DaggerWebDash` but it would be much easier to parse if there was a complete tutorial example of a workflow and/or some links to example usage in the Julia community.

seabbs:
To add to this ideally we would be able to visualise the DAG both before and after running it - is that supported? 

Slightly unrelated but are you aware of any real world uses cases of relatively complex Dagger workflows? We have had a look but not been very successful. We are particularly interested in modelling our pipeline (and the dependencies between jobs) and being able to abstract parallelization without pipeline changes (so we could switch from local -> Slurm -> something cloud like azure batch) so resources in that direction would be great.

jpsamaroo:
Hey there! There is definitely a way to do this, but I realize the full example isn't documented. You can do something like this:

```julia
using Dagger, GraphViz

# Enable logging, with required logging events for GraphViz
Dagger.enable_logging!(;tasknames=true, taskdeps=true, taskargs=true, taskargmoves=true)

# Run your code that uses `Dagger.@spawn` (or use a `DArray` or `DTable`)
x = Dagger.@spawn 1+1
y = Dagger.@spawn x*2
z = Dagger.@spawn x/3
fetch(z)

# Fetch and show the logs (shows up automatically in Jupyter)
logs = Dagger.fetch_logs!()
Dagger.render_logs(logs, :graphviz)
```

However, I've noticed that the above example doesn't produce any viewable DAG, because there was an oversight in the dependency calculation that missed some task dependencies - I'm working on fixing this now, along with writing docs for all of this! There are also a number of options for the `:graphviz` renderer that are passed as keyword arguments - I'll document these shortly. For now, if you do `Dagger.render_logs(logs, :graphviz; disconnected=true)`, you can at least see each individual task as a bubble, but once I get my fixes merged, you'll also be able to see the dependencies between tasks.

SamuelBrand1:
Thanks for getting back to us so quickly!

Nice idea to build in a dep with `GraphViz`, look forward to the fix!

[...]

PR #531, opened 2024-06-16
Fix write dag

SmalRat:
Fixes for write_dag() and show_logs() functions. More details on #512 .

### DIFF
--- a/ext/GraphVizSimpleExt.jl
+++ b/ext/GraphVizSimpleExt.jl
@@ -33,6 +33,22 @@ function node_id(t::Chunk)
     _part_labels[t]
 end
 
+function node_name(t::Thunk)
+    "n_$(t.id)"
+end
+
+function node_name(t::Chunk)
+    _part_labels[t]
+end
+
+function node_name(name::String)
+    "n_$name"
+end
+
+function node_name(id)
+    "n_$id"
+end
+
 function write_dag(io, t::Thunk)
     !istask(t) && return
     deps = dependents(t)
@@ -43,7 +59,7 @@ function write_dag(io, t::Thunk)
     for (k, v) in deps
         for dep in v
             if isa(k, Union{Chunk, Thunk})
-                println(io, "$(node_id(k)) -> $(node_id(dep))")
+                println(io, "$(node_name(k)) -> $(node_name(dep))")
             end
         end
     end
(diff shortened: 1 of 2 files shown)


jpsamaroo:
Awesome stuff! I'm happy to accept this as-is if you'd like, but there are also a few things we can improve if you want to tackle them:
- Add a `render_logs` alternative (similar to `render_logs(logs, :graphviz)`) that returns a renderable GraphViz object (and equivalently, we could implement `show_logs(logs, :graphviz)` like you've done here for rendering directly to an `IO`)
- Consider adding a `save_logs` function that drives Cairo to save renderings to a file
- Document (under `docs/src/logging-visualization.md`) this newly-supported renderer, so that users can more easily discover it

Let me know if you want this merged now or if you want to try to tackle some of these, I can always add these to the TODO list :smile:

SmalRat:
> * Add a `render_logs` alternative (similar to `render_logs(logs, :graphviz)`) that returns a renderable GraphViz object (and equivalently, we could implement `show_logs(logs, :graphviz)` like you've done here for rendering directly to an `IO`)
> * Consider adding a `save_logs` function that drives Cairo to save renderings to a file, or make `render_logs(path::String, ...)` somehow dispatch to do this (maybe via FileIO.jl)
> * Document (under `docs/src/logging-visualization.md`) this newly-supported renderer, so that users can more easily discover it
> 
> Let me know if you want this merged now or if you want to try to tackle some of these, I can always add these to the TODO list 😄

Looks like it is not a problem to add this. I will do that soon

SmalRat:
Hi!

There is an issue with implementing these features:

> * Add a render_logs alternative (similar to render_logs(logs, :graphviz)) that returns a renderable GraphViz object (and equivalently, we could implement show_logs(logs, :graphviz) like you've done here for rendering directly to an IO)

> * Consider adding a save_logs function that drives Cairo to save renderings to a file, or make render_logs(path::String, ...) somehow dispatch to do this (maybe via FileIO.jl)

It seems that GraphVizSimpleExt, which I am working on, was intended to be used when GraphViz is not imported (with GraphVizExt for the opposite case). Am I correct in this assumption?

If that is the case, GraphVizSimpleExt can not return a renderable GraphViz object, which we want from the `render_logs` function and which is to be used in `save_logs`.

jpsamaroo:
Hi, I'm sorry about the long delay in my reply! You are welcome to add GraphViz as a dependency to `GraphVizSimpleExt` - I just didn't add it since I hadn't gotten the code really working, and it didn't currently depend on it. But it definitely seems like the right thing to do!

PR #532, opened 2024-06-17
Dead worker handling

JamesWrigley:
I've cherry-picked two commits from the streaming branch that improve handling of dead workers, @mofeing this should fix the errors that you're seeing (in particular 1fa2ea4). Originally added in #496.


jpsamaroo:
Thanks again!

PR #540, opened 2024-07-02
Add render/show_logs to chrome_trace

m-fila:
Chrome tracing is a fairly popular tool for working with traces and process timelines. The traces can be not only displayed but also analyzed and queried with SQL

The Dagger logs can be transformed to a [chrome-trace schema](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.yr4qxyxotyw) json format. The resulting json file can be loaded to https://ui.perfetto.dev/ or legacy `about:tracing` in a chrome-based browser

```jl
Dagger.enable_logging!(timeline=true)
...
logs = Dagger.fetch_logs!()
open("chr_trace.json", "w") do io
    Dagger.show_logs(io, logs, :chrome_trace)
end
```

This can be considered an alternative to the in-house `:plots_gantt`

---

Example output ([json file](https://github.com/user-attachments/files/16063094/chr_trace.json)):

perfetto

For comparison, the same logs printed with `:plots_gantt`

gantt

---

Optional dependency on `JSON3` is added, but it's used only to get a json-string from a Dict so it should be possible to easily replace it with any other JSON library in case you have a preference

For now only `:compute` events are used (same as in `:plots_gantt`) but it should be possible to add support for other types of evetns too


jpsamaroo:
Sorry for the delay - this is really cool! I'll review this soon, it looks already really good :heart:

docs/src/logging-visualization.md
```diff
@@ -20,6 +20,7 @@ converted to a `Val` for dispatch purposes
 Built-in rendering support exists for:
 - `render_logs(logs, :graphviz)` to generate a graph diagram of executed tasks and their dependencies
 - `render_logs(logs, :plots_gantt)` to generate a Gantt chart of task execution across all processors
+- `render_logs(logs, :chrome_trace)` to generate a task execution timeline in the chrome-trace format (view in [perfetto web UI](https://ui.perfetto.dev/) or `about:tracing` in a chrome-based browser)
```
jpsamaroo, line 23: Would you also be willing to add a section above for the `show_logs` method you added? It can just refer to this description, or vice versa.

ext/JSON3Ext.jl
```diff
@@ -0,0 +1,54 @@
...
+            execution_logs[tid][:ts] = t_start
+            execution_logs[tid][:dur] = t_stop - t_start
+            execution_logs[tid][:pid] = proc.owner
+            execution_logs[tid][:tid] = proc.tid # thread id
+        elseif category == :add_thunk
+            tid = logs[w][:id][start_idx].thunk_id
+            if !haskey(execution_logs, tid)
+                execution_logs[tid] = Dict{Symbol,Any}()
+            end
+            f = logs[w][:timeline][start_idx].f
```
jpsamaroo, line 32: Instead of using `:timeline` (which is preferably avoided, as it captures every single object submitted to Dagger, and can cause memory to leak), you can use `:tasknames`, which we use in GraphVizExt to get nice function names. I typically do `first(split(name, ' '))` to get just the function name, as the name is in the format `"fn [tid]"`.

ext/JSON3Ext.jl
```diff
@@ -0,0 +1,54 @@
...
+            if !haskey(execution_logs, tid)
+                execution_logs[tid] = Dict{Symbol,Any}()
+            end
+            t_start = logs[w][:core][start_idx].timestamp / 1e3 # us
+            t_stop = logs[w][:core][finish_idx].timestamp / 1e3 # us
+            proc = logs[w][:id][start_idx].processor
+            execution_logs[tid][:ts] = t_start
+            execution_logs[tid][:dur] = t_stop - t_start
+            execution_logs[tid][:pid] = proc.owner
+            execution_logs[tid][:tid] = proc.tid # thread id
```
jpsamaroo, line 26: One thing to beware of - only `ThreadProc` has `tid`; the GPU processors (like `CuArrayDeviceProc`) does not have this field. Similarly to how we have `Dagger.root_worker_id(::Processor)` to get the Distributed worker that a processor is associated with (essentially the `owner` field), we could also have a helper like `Dagger.processor_thread_id(::Processor)` which would return `tid` for `ThreadProc`, and something like `device_id` for GPU processors.

You are welcome to ignore this or put it off for later - I will also happily add it when I get around to it. If you want to do the easy thing, you can also just skip any events where `!(proc isa Dagger.ThreadProc)`, and maybe just emit a warning that events were skipped? It's totally up to you!

jpsamaroo (commented):
Wonderful work! I think I'll plan to recommend this when users submit bugs/performance issues.

Just a few desired tweaks, and I'd be happy to merge this :smile:

ext/JSON3Ext.jl
m-fila: For now I'll put all the non-`ThreadProc` as a single "unknown" pid/tid and print warning. I haven't used Dagger with GPU yet, but I plan to, so maybe I'll add a revision then

docs/src/logging-visualization.md
jpsamaroo: Perfect, thanks!

jpsamaroo (approved):

jpsamaroo:
Thanks so much @m-fila !

PR #541, opened 2024-07-04
Include datadeps in API documentation

m-fila:
The datadeps (`spawn_datadeps`, `In`, `Out`, etc.) have their own usage-page in the docs but weren't mention in the API documentation. The already provided docstrings can be used

I think it would be also useful to add a remark in the documentation that `spawn_datadeps` waits for all the tasks spawned. This part was unclear for me and I had to check the source code

### DIFF
--- a/docs/src/api-dagger/functions.md
+++ b/docs/src/api-dagger/functions.md
@@ -30,6 +30,11 @@ mutable
 shard
 ```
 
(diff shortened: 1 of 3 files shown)


jpsamaroo (approved):
Ahh wonderful, thanks for catching these!

PR #546, opened 2024-07-11
Add rand! implementation

pszufe:


jpsamaroo:
Switching to draft while I finish testing some new datadeps infrastructure to detect and skip non-written arguments (among other related things).

src/array/operators.jl
```diff
@@ -111,3 +111,24 @@ end
 
 Base.first(A::DArray) = A[begin]
 Base.last(A::DArray) = A[end]
+
+# In-place operations
+
+function Base.map!(f, a::DArray{T}) where T
+    Dagger.spawn_datadeps() do
+        for ca in vec(chunks(a))
```
JamesWrigley, line 119: Out of curiosity, is it necessary to flatten `chunks(a)` like this?

src/array/random.jl
```diff
@@ -0,0 +1,30 @@
+using Random
```
JamesWrigley, line 1: Nitpicking a bit, could we do explicit imports instead of `using`? It makes a bit easier to see where symbols come from.

src/array/random.jl
```diff
@@ -0,0 +1,30 @@
...
+
+randfork(rng::R, n::Integer) where {R<:AbstractRNG} =
+    R(rand(copy(rng), Int) + n)
+
+function Random.rand!(rng::AbstractRNG, A::DArray{T}) where T
+    part_sz = prod(map(length, first(A.subdomains).indexes))
+    Dagger.spawn_datadeps() do
+        for Ac in chunks(A)
+            rng = randfork(rng, part_sz)
+            new_rng = copy(rng)
```
JamesWrigley, line 14: Is it necessary to copy `rng` again here since `randfork()` creates a new one anyway? (same thing for the `randn!()` definition below)

JamesWrigley (changes requested):

src/array/operators.jl
jpsamaroo: Good point! It is not necessary.
```suggestion
        for ca in chunks(a)
```

src/array/random.jl
jpsamaroo: Sure, and I'll pull this one out to the top-level `Dagger.jl` file

src/array/random.jl
jpsamaroo: I had this here because we may pass the `rng` argument as-is to the task, and if it's executed on a local thread, we want to make sure that it won't alias with future uses of `rng` within this function. But datadeps tasks don't start until the end of the block, so it should be fine to not copy it again.

jpsamaroo:
There appears to be some significant segression or hang in datadeps making the tests run too long, which I need to resolve before merging this.

jpsamaroo:
Thanks @pszufe for the implementation and @JamesWrigley for the review!

Issue #548, opened 2024-07-13
Parallelization across threads does not appear to work correctly


wolthom:
## Context
In the following minimal example, an outer function executes the inner function many times in parallel and is itself executed multiple times. The problem is trivially parallelizable, i.e. there are no dependencies across tasks.

I'm running Julia with 22 threads:
```julia-repl
julia> Base.Threads.nthreads()
22
```

## Copy-pasteable example
```julia
using Dagger

# Inner function
function inner_sleep()
    sleep(2)
end

# Outer function
function outer_sleep()
    @sync for _ in 1:10
        Dagger.@spawn inner_sleep()
    end
    Base.Threads.threadid()
end

# Run outer task N times in parallel
@time fetch.([Dagger.@spawn outer_sleep() for _ in 1:2])
```

## Results
If my understanding of the task-based parallelism approach is correct, this should take roughly 2 seconds regardless of how many outer tasks I'm spawning.

This actually takes ~4 seconds instead, on my machine.

jpsamaroo:
You'll want to let Dagger execute multiple of `inner_sleep` on each thread at a time with this for each call:

`Dagger.@spawn occupancy=Dict(Dagger.ThreadProc=>0) inner_sleep()`

PR #549, opened 2024-07-13
add documentation for Task / Thread occupancy

wolthom:
This PR adds a section to the Task Spawning documentation page to explain the behavior observed in #548.

### DIFF
--- a/docs/src/task-spawning.md
+++ b/docs/src/task-spawning.md
@@ -27,6 +27,11 @@ its result will be passed into the function receiving the argument. If the
 argument is *not* an [`DTask`](@ref) (instead, some other type of Julia object),
 it'll be passed as-is to the function `f` (with some exceptions).
 
+!!! note "Task / thread occupancy"
+    By default, `Dagger` assumes that tasks saturate the thread they are running on and does not try to schedule other tasks on the thread.
+    This default can be controlled by specifying [`Sch.ThunkOptions`](@ref) (more details can be found under [Scheduler and Thunk options](@ref)).\
+    The section [Changing the thread occupancy for low-utilization tasks](@ref) shows a runnable example of how to achieve this.
+
 ## Options
 
 The [`Options`](@ref Dagger.Options) struct in the second argument position is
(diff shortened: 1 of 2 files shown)


docs/src/task-spawning.md
```diff
@@ -27,6 +27,11 @@ its result will be passed into the function receiving the argument. If the
 argument is *not* an [`DTask`](@ref) (instead, some other type of Julia object),
 it'll be passed as-is to the function `f` (with some exceptions).
 
+!!! note "Task / thread occupancy"
+    By default, `Dagger` assumes that tasks saturate the thread they are running on and does not try to schedule other tasks on the thread.
+    This default can be controlled by specifying [`Sch.ThunkOptions`](@ref) (more details can be found under [Scheduler and Thunk options](@ref)).\
```
jpsamaroo, line 32:
```suggestion
    This default can be controlled by specifying [`Sch.ThunkOptions`](@ref) (more details can be found under [Scheduler and Thunk options](@ref)).
```

I think that was a typo?

jpsamaroo (commented):
Looks great! I had trouble running it locally - Documenter hangs at "ExpandTemplates: expanding markdown templates" - any ideas why?

wolthom:
@jpsamaroo I think I may have an idea:

Did you run the `make.jl` script from the commandline? E.g. via `julia --threads=12 --project=. make.jl`?

For some reason that hangs for me as well. However, if I start a REPL, activate the docs environment and `include("make.jl")`, it builds normally for me.

That also explains why the Documentation CI step timed out. (Just realized that).

Could this potentially be an interaction between the Dagger scheduler / the main thread of the REPL / ...?

jpsamaroo:
Yeah I ran it from the CLI as well. I'm running with the `JULIA_DEBUG=Dagger` env var set, but nothing prints - this makes me think that either Documenter is filtering it out. I forced it in Dagger directly, yet it still doesn't show anything, so probably it's being redirected into a file/pipe for output. This makes things particularly hard to debug... maybe I'll need to implement a debug helper that sends output to a file or something?

wolthom:
Interestingly enough, I do get output when running from the CLI with `JULIA_DEBUG=Dagger`:
image

If this is specific to this environment / setup and not some underlying actual bug, I could also turn this from executed code to "dumb" code snippets as a workaround.

jpsamaroo:
@wolthom yeah let's do that for now - I don't have a whole lot of time to investigate this right now, and I don't want to block adding this information to the docs, so let's just use "dump" code blocks, and maybe you can just include the result from running on your local system?

wolthom:
@jpsamaroo I have updated the `task-spawning.md` note as discussed above. Now the code blocks should no longer be executed as part of the documentation build (and thus not hang).

Could you verify if this also works as expected for you?

jpsamaroo:
Worked great, thank you!

PR #551, opened 2024-07-13
docs: Create favicon and add it to the documentation

wolthom:
This PR combines three changes:
1. The icons are shifted to `docs/src/assets` (mirroring other well-documented packages I've checked)
2. A new icon without text and the corresponding `favicon.ico` are created
3. `make.jl` is updated to use the `favicon.ico`

By doing this, the documentation shows the Dagger icon in the upper left corner and as the favicon in the browser tab bar:
image


jpsamaroo:
Amazing, thank you! This looks much better now :smile:

PR #559, opened 2024-07-22
Replace all uses of @async with Threads.@spawn

JamesWrigley:
~~- Use `errormonitor_tracked()` for executing `DTask`'s~~
~~- Set `allow_errors=false` by default in the eager scheduler.~~

~~I don't know why we had `allow_errors=true` originally but it seems way too permissive, it caused some internal exceptions to not be logged when I was working on the streaming branch.~~

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -40,7 +40,7 @@
         for (name, t) in tracked
             if t.state == :runnable
                 @warn "Waiting on $name"
-                @async Base.throwto(t, InterruptException())
+                Threads.@spawn Base.throwto(t, InterruptException())
             end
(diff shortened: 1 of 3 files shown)


JamesWrigley:
In hindsight I should've run the tests locally first :laughing: I'll try to fix those errors.

JamesWrigley:
Using `errormonitor_tracked()` would still print out some ugly error messages when I don't think we want it to so I decided to revert that, now the only change is to use `Threads.@spawn` everywhere instead of `@async`.

jpsamaroo:
Thanks!

PR #566, opened 2024-08-16
Make Dagger.cancel!() handle a shutting down/shutdown scheduler

JamesWrigley:
Cherry-picked from #463.


jpsamaroo:
Thanks for catching this!

PR #568, opened 2024-08-18
Initial support for robustly migrating streaming tasks

JamesWrigley:
This works by converting the output buffers into a safely-serializeable container and sending that to the new node.

(needs some tests though)


jpsamaroo (approved):

jpsamaroo:
Thanks @JamesWrigley ! We're just about ready to merge the whole thing :partying_face:

PR #569, opened 2024-08-19
add `show_logs` for `:graphviz` 

m-fila:
While `render_logs` graphviz creates a `GraphViz` object that can be later saved to a picture sometimes it's more convenient to be able to save to a `dot` text format (tune the file if needed, for instance remove some details or vertices for presentation) and then render to a picture with an external `dot` tool.

Both `show_logs` and `render_logs` for `:graphviz` use the same implementation. The `show_logs` and associated functions were moved to visualizations from extension as they don't depend on `GraphViz` package (`render_logs` does)


jpsamaroo:
Thanks! I'm not going to merge this just yet, as I'd like to integrate it with #560 first. I'll recreate the commits with you as the author.

jpsamaroo:
Implemented in #560

PR #570, opened 2024-08-27
use user given task names in chrome trace

m-fila:
Adding user defined task names #560  in visualization with `:chrome-trace` #540 . By default an automatically inferred named is used, if a task has user provided name then it will be used instead

`logs_event_pairs` is not used anymore as the `:data_annotation` are not present in it

### DIFF
--- a/ext/JSON3Ext.jl
+++ b/ext/JSON3Ext.jl
@@ -13,69 +13,56 @@ function logs_to_chrome_trace(logs::Dict)
     tid_to_uid = Dict{Int,UInt}()
     uid_to_name = Dict{UInt,String}()
     add_unknown_procs_metadata = false
-    for w in keys(logs)
-        for idx in 1:length(logs[w][:core])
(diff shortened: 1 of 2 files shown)


jpsamaroo:
> `logs_event_pairs` is not used anymore as the `:data_annotation` are not present in it

That's a bug, I did something stupid by making the IDs of `:data_annotation` start/finish not match up (which `logs_event_pairs` rightly skips over, since it can't then provide a matching pair). Making the ID field in https://github.com/JuliaParallel/Dagger.jl/blob/1a317c38cace53008154d860eef11d497cda3273/src/utils/logging.jl#L140 match the `timespan_start` line above it would fix this issue.

m-fila:
Thanks 😃  I found that mismatch but then read the comment   `# TODO: Remove redundant log event` and fought it's intended. I'll fix and update the PR tomorrow

jpsamaroo:
Awesome, thank you!

PR #576, opened 2024-10-11
Use `[sources]` to specify local deps for documentation

### LINKED ISSUE #552: Dagger documentation does not build locally
While working on a PR to the documentation, I have encountered the issue that the documentation does not build without local changes.

Specifically, I get the following error:
image

This can be worked around by removing the `@docs` marker in the corresponding code snippet in `src/api-timespanlogging/functions.md` (without committing this change, of course).
In CI this does appear to build successfully due to locally dev'ed packages(?).

Is there a good way to fix this?

m-fila:
In julia 1.11 new section `[sources]` was added allowing for specifying dependencies to local or unregistered packages.
This can be used in the `docs` to specify dependencies to local packages, so the documentation can be build locally without extra tweaks

Closes #552

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -21,7 +21,7 @@ jobs:
           version: '1'
       - uses: julia-actions/cache@v2
       - name: Install dependencies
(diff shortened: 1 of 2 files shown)


JamesWrigley (approved):
We can wait for CI to finish before merging, but LGTM!

jpsamaroo:
Awesome stuff, thanks @m-fila !

PR #578, opened 2024-10-23
Update CI jobs

JamesWrigley:
Now we always use the latest stable and nightly builds, except for the 1.9 tests.

I also noticed that our codecov hasn't been updated for a couple years: https://app.codecov.io/gh/JuliaParallel/Dagger.jl/commits
Now that codecov doesn't allow tokenless uploading anymore I think we have to set up the `SECRET_CODECOV_TOKEN` as described here: https://github.com/JuliaGPU/buildkite?tab=readme-ov-file#using-secrets
(an admin will need to do that since I don't have access to the token)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -25,27 +25,30 @@ steps:
           julia_args: "--threads=1"
       - JuliaCI/julia-coverage#v1:
           codecov: true
-  - label: Julia 1.10
+
+  - label: Julia 1
     timeout_in_minutes: 90
     <<: *test
     plugins:
       - JuliaCI/julia#v1:
-          version: "1.10"
+          version: "1"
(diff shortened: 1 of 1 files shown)


JamesWrigley:
The nightly failure is real, I'm guessing we need to merge this: https://github.com/JuliaConcurrent/UnsafeAtomics.jl/pull/14

jpsamaroo:
Tokenless uploads should become available once https://github.com/JuliaCI/julia-coverage-buildkite-plugin/pull/12 is merged, so I'm going to merge this. Also, https://github.com/JuliaConcurrent/UnsafeAtomics.jl/pull/14 got merged, but we're still waiting on a new release - I think it's worth updating CI in the meantime, even if nightly is failing for now.

jpsamaroo:
Thanks!

PR #579, opened 2024-11-16
Make `Dagger.finish_stream()` propagate downstream

JamesWrigley:
Previously a streaming task calling `Dagger.finish_stream()` would only stop the caller, but now it will also stop all downstream tasks. This is done by:
- Getting the output handler tasks to close their `RemoteChannel` when exiting.
- Making the input handler tasks close their buffers when the `RemoteChannel` is closed.
- Exiting `stream!()` when an input buffer is closed.

One question, what exactly are the `DropBuffer` tests for? Are they just for testing `DropBuffer` (its usefulness is not clear to me), or to test that streaming tasks should throw exceptions when `fetch()`'d after cancelling? For now I've just disabled those tests because with the new behaviour a task stopping for _any reason_ will also stop the downstream tasks. So currently in the tests the `x` task will reach `max_evals` and finish, which also cause `A` to finish, which means everything stops gracefully and `A` is never cancelled so it doesn't throw an exception.


jpsamaroo:
I'm cherry-picking portions of this PR into https://github.com/JuliaParallel/Dagger.jl/pull/463. I'll also be implementing a `waitany`-based DAG teardown feature that should accomplish the same as this PR's intention but also propagating upstream too.

Thanks for this!

JamesWrigley:
Sounds good :) I'll close this then.

PR #580, opened 2024-11-25
datadeps: Don't skip copy on no writedep

jpsamaroo:
It's incorrect to skip on just this, because we might still need to copy for the purpose of adapting types, e.g. CPU -> GPU.


jpsamaroo:
Julia 1 failures are unrelated

PR #581, opened 2024-12-01
Streaming thread safety

JamesWrigley:
There's a few changes in here, I would recommend reviewing each commit individually. The most important one is 96a0c46e.

Side note: I see that we have both `take!(::StreamStore, ::UInt)` and `take!(::StreamingValue)`. The former is unused AFAICT, and it differs from the latter in that it will wait for the store to be notified when empty instead of spinning. Should we use it instead? Or delete it?


JamesWrigley:
Blegh, evidently didn't get all of them: https://buildkite.com/julialang/dagger-dot-jl/builds/1394#01937fb0-b82c-45c0-b5b0-02471b806bab/436-1441

jpsamaroo:
Most of these changes were cherry-picked or manually ported into #463, which is generally passing CI, so I'm going to close this. Thanks for putting these fixes together!

PR #586, opened 2025-01-13
Add MPI integration to datadeps

fda-tome:
MPI suite and base implementations such as processors and messaging with P2P MPI directives that fit into both the scheduler logic and datadeps scheme.


jpsamaroo:
Implemented in #720

PR #587, opened 2025-01-25
Hash `USER` environment variable during system_uuid fallback

nHackel:
I recently encountered a system where the `USER` environment variable got set to "domain/user". Such a name then gets interpreted as a filepath which breaks the `system_uuid_fallback` logic.

As a workaround this PR hashes the variable before using it in the filepath creation. Same user names should still result in the same hash values


nHackel:
Ah actually the docs mention that the hash may change when a new process is started. I'm not entirely sure if that is an issue here, because the workers should be started before Dagger is loaded according to the docs. I also wasn't able to locally produce different hashes for the same value

jpsamaroo:
Thanks!

PR #590, opened 2025-03-09
docs: writes to parameter passed to Out(), doesnt read from it

n0rbed:
as the title states, `C` is written to not read from


jpsamaroo:
Great catch, thank you!

PR #601, opened 2025-06-05
tests: Test show_logs method

jpsamaroo:


jpsamaroo:
Apparently we already had this method, whoops...

PR #625, opened 2025-07-01
stencils: Add docs and tests

jpsamaroo:
From jules.google.com


jpsamaroo:
Already included in #564

PR #630, opened 2025-07-08
Enhance DArray Distribution with Processor Assignment

AkhilAkkapelli:
- Updated Distribute struct to include a processor grid for better control over data distribution.
- Modified constructors for Distribute to accept processor grid as an optional parameter.
- Enhanced distribute functions to support various assignment strategies: :arbitrary, :blockrow, :blockcol, :cyclicrow, and :cycliccol.
- Added tests for new assignment strategies, ensuring correct processor allocation and chunking for DArray, DVector, and DMatrix.
- Improved test coverage for both auto and explicit block distributions with processor assignments.

### DIFF
--- a/docs/src/darray.md
+++ b/docs/src/darray.md
@@ -211,6 +211,254 @@ across the workers in the Julia cluster in a relatively even distribution;
 future operations on a `DArray` may produce a different distribution from the
 one chosen by previous calls.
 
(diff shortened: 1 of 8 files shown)


jpsamaroo:
Thanks again @AkhilAkkapelli !

PR #639, opened 2025-07-29
DTask/Sch: Optimize DTask futures, optimize proc in scope checks

jpsamaroo:


jpsamaroo:
Seems like the `LocalFuture` changes are having some concurrency issues, presumably due to the shenanigans we perform around `ThunkFuture` serialization/deserialization. The other commit was already merged, so this PR is on the back-burner for now.

PR #640, opened 2025-07-30
scopes/Sch: Add optimized check for proc in scope

jpsamaroo:
Split from #639, to see how just this commit performs


jpsamaroo:
Yay for easy win!

Issue #647, opened 2025-08-07
3D FFT gets `DivideError: integer division error`


islent:
My MWE:
```jl
using Distributed
addprocs(8)
@everywhere using Dagger

using FFTW
N = 64
A = rand(ComplexF64,N,N,N);
DA = DArray(A);
fft(DA)
```

It throws error:
```jl
ERROR: DivideError: integer division error
Stacktrace:
  [1] div
    @ .\int.jl:295 [inlined]
  [2] _cumlength(len::Int64, step::Int64)
    @ Dagger E:\.julia\packages\Dagger\Jl5aj\src\array\alloc.jl:62
  [3] map
    @ .\tuple.jl:383 [inlined]
  [4] map
    @ .\tuple.jl:386 [inlined]
  [5] partition
    @ E:\.julia\packages\Dagger\Jl5aj\src\array\alloc.jl:69 [inlined]
  [6] zeros(p::Blocks{3}, eltype::Type, dims::Tuple{Int64, Int64, Int64}; assignment::Symbol)
    @ Dagger E:\.julia\packages\Dagger\Jl5aj\src\array\alloc.jl:153
  [7] zeros
    @ E:\.julia\packages\Dagger\Jl5aj\src\array\alloc.jl:151 [inlined]
  [8] _fft!(output::DArray{…}, input::DArray{…}, dims::Tuple{…}; decomp::AbstractFFTsExt.Pencil)
    @ AbstractFFTsExt E:\.julia\packages\Dagger\Jl5aj\ext\AbstractFFTsExt.jl:124
  [9] fft(DA::DArray{ComplexF64, 3, Blocks{3}, typeof(cat)}, dims::Tuple{Int64, Int64, Int64}; decomp::AbstractFFTsExt.Pencil)
    @ AbstractFFTsExt E:\.julia\packages\Dagger\Jl5aj\ext\AbstractFFTsExt.jl:62
 [10] fft
    @ E:\.julia\packages\Dagger\Jl5aj\ext\AbstractFFTsExt.jl:59 [inlined]
 [11] fft(DA::DArray{ComplexF64, 3, Blocks{3}, typeof(cat)})
    @ AbstractFFTsExt E:\.julia\packages\Dagger\Jl5aj\ext\AbstractFFTsExt.jl:59
 [12] top-level scope
    @ REPL[30]:1
```

I also tried to manually control the partitions, whereas the `DivideError` cannot be solved:
```jl
[...]

PR #648, opened 2025-08-08
change `div` to `cld` to avoid DivideError in FFT

islent:
Solves https://github.com/JuliaParallel/Dagger.jl/issues/647


islent:
Yes, the MWE works for this branch!

PR #687, opened 2026-02-23
CI: Add multi-threaded job

jpsamaroo:
Let's see if we can speed up CI by giving Dagger some threads to work with.


jpsamaroo:
Implemented in #706

PR #689, opened 2026-02-27
Free GPU memory eagerly, and various linalg fixes

jpsamaroo:
We now use `unsafe_free!` to eagerly free Datadeps and DArray GPU allocations when safe to do so.

This PR also:
- Fixes LU for GPUs
- Fixes Cholesky for `DVector` inputs
- Adds `+` and `-` methods for `DArray`s
- Fixes a scheduler bug around error reporting in large task graphs
- Adapts `RefValue` to a new `GPURef`, required for LU and other algorithms

Todo:
- [x] Add `unsafe_free!` to stencils
- [x] Evaluate if QR/CAQR can use `unsafe_free!`
- [x] Add CPU/GPU tests for Cholesky with `DVector` inputs
- [x] Add GPU tests for Cholesky and LU
- [x] Add GPU tests for `+` and `-`
- [ ] Add BLAS alternatives for GPU backends which don't have BLAS wired up

### DIFF
--- a/ext/CUDAExt.jl
+++ b/ext/CUDAExt.jl
@@ -335,6 +335,37 @@ for lib in [BLAS, LAPACK]
     end
 end
 
+# Adapt RefValue
+Dagger.move(from_proc::CPUProc, to_proc::CuArrayDeviceProc, x::Base.RefValue) =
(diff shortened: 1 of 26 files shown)


jpsamaroo:
Thanks for the help @fda-tome !

PR #697, opened 2026-04-12
Make transfer rate metric per-processor in scheduler

riteshbhirud:
The transfer rate metric in the scheduler was a single global `Ref{UInt64}`, 
unlike other metrics like `worker_time_pressure` which are per-processor. 
This expands `transfer_rate` to `worker_transfer_rate::Dict{Int,Dict{Processor,UInt64}}` 
following the same pattern, so each processor accumulates its own observed transfer rate 
rather than averaging across all processors.


jpsamaroo:
Thanks so much!

PR #700, opened 2026-04-17
Replace CUDA.synchronize() with CuEvent-based synchronization

victorcamaraa:
Replaces all `CUDA.synchronize()` device-wide barriers in `CUDAExt.jl` with `CuEvent`-based synchronization.

The old approach issued a full `cuDeviceSynchronize` on every data movement and `gpu_synchronize()` call, stalling the CPU until the entire device was idle. The new approach records a `CuEvent` on the relevant stream and issues a GPU-side wait, so only the streams that actually share data are synchronized and the CPU is never blocked unnecessarily.

The one exception is DtoH transfers, where the CPU genuinely needs to wait for data to arrive in host memory, those use `CUDA.synchronize(ev)`, which is a CPU wait scoped to a single event rather than the full device.

### Changes
- `_sync_with_context`: captures caller's stream before context switch, replaces `CUDA.synchronize()` with `CuEvent` record + GPU-side `CUDA.wait`
- `gpu_synchronize`: captures user's ambient stream before context switch, same event pattern
- HtoD `move`: drops `CUDA.synchronize()` entirely — stream ordering is sufficient
- DtoH `move`: replaces full barrier with `CUDA.synchronize(ev)` scoped to the active stream
- DtoD same-device `move`: drops barrier — task graph ordering guarantees upstream completion
- DtoD cross-device `move`: replaces source-side barrier with a `CuEvent` that the destination stream GPU-waits on
- `gpu_synchronize(::Val{:CUDA})`: refactored to call `gpu_synchronize(proc)` per device for consistency

#### Notes:
- These implementations were tested on a single GTX 1060 6GB; therefore, I was not able to test DtoD
- Debugging and partial implementation were assisted by Claude Sonnet 4.7 and Gemini

### DIFF
--- a/ext/CUDAExt.jl
+++ b/ext/CUDAExt.jl
@@ -16,7 +16,7 @@ if isdefined(Base, :get_extension)
 else
     import ..CUDA
 end
-import CUDA: CuDevice, CuContext, CuStream, CuArray, CUDABackend
+import CUDA: CuDevice, CuContext, CuStream, CuArray, CUDABackend, CuEvent
 import CUDA: devices, attribute, context, context!, stream, stream!
 import CUDA: CUBLAS, CUSOLVER
 
@@ -103,8 +103,11 @@ function with_context(f, x)
 end
 
 function _sync_with_context(x::Union{Dagger.Processor,Dagger.MemorySpace})
+    caller_stream = stream()
     with_context(x) do
-        CUDA.synchronize()
+        ev = CUDA.CuEvent()
+        CUDA.record(ev, stream())
+        CUDA.wait(ev, caller_stream)
     end
 end
 function sync_with_context(x::Union{Dagger.Processor,Dagger.MemorySpace})
@@ -164,7 +167,6 @@ function Dagger.move(from_proc::CPUProc, to_proc::CuArrayDeviceProc, x::Chunk)
     cpu_data = remotecall_fetch(unwrap, from_w, x)
     with_context(to_proc) do
         arr = adapt(CuArray, cpu_data)
-        CUDA.synchronize()
         return arr
     end
 end
@@ -175,7 +177,6 @@ function Dagger.move(from_proc::CPUProc, to_proc::CuArrayDeviceProc, x::CuArray)
     with_context(to_proc) do
         _x = similar(x)
         copyto!(_x, x)
-        CUDA.synchronize()
         return _x
     end
 end
@@ -184,9 +185,7 @@ end
 function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CPUProc, x)
     with_context(from_proc) do
         CUDA.synchronize()
-        _x = adapt(Array, x)
-        CUDA.synchronize()
-        return _x
+        return adapt(Array, x)
     end
 end
 function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CPUProc, x::Chunk)
@@ -203,7 +202,6 @@ function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CPUProc, x::CuArray{
         CUDA.synchronize()
         _x = Array{T,N}(undef, size(x))
         copyto!(_x, x)
-        CUDA.synchronize()
         return _x
     end
 end
@@ -211,18 +209,22 @@ end
 # Out-of-place DtoD
 function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CuArrayDeviceProc, x::Dagger.Chunk{T}) where T<:CuArray
     if from_proc == to_proc
-        # Same process and GPU, no change
-        arr = unwrap(x)
-        with_context(CUDA.synchronize, from_proc)
-        return arr
+        # Same process and GPU, no change.
+        # Stream ordering guarantees safety; no sync needed.
+        return unwrap(x)
+        
     elseif Dagger.root_worker_id(from_proc) == Dagger.root_worker_id(to_proc)
         # Same process but different GPUs, use DtoD copy
         from_arr = unwrap(x)
-        with_context(CUDA.synchronize, from_proc)
+        ev = CUDA.CuEvent()
+        with_context(from_proc) do
+            CUDA.record(ev, stream())
+        end
+        
         return with_context(to_proc) do
+            CUDA.wait(ev, stream()) 
             to_arr = similar(from_arr)
             copyto!(to_arr, from_arr)
-            CUDA.synchronize()
             return to_arr
         end
     elseif Dagger.system_uuid(from_proc.owner) == Dagger.system_uuid(to_proc.owner) && from_proc.device_uuid == to_proc.device_uuid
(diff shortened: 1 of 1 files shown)


ext/CUDAExt.jl
```diff
@@ -390,13 +412,21 @@ Dagger.gpu_kernel_backend(::CuArrayDeviceProc) = CUDABackend()
...
     CUDA.device!(f, proc.device)
 function Dagger.gpu_synchronize(proc::CuArrayDeviceProc)
+    user_stream = stream()
+
     with_context(proc) do
-        CUDA.synchronize()
+        ev = CUDA.CuEvent()
+        CUDA.record(ev, stream()) 
+        CUDA.wait(ev, user_stream)
+
```
jpsamaroo, line 408:
```suggestion
```

ext/CUDAExt.jl
```diff
@@ -175,18 +176,18 @@ function Dagger.move(from_proc::CPUProc, to_proc::CuArrayDeviceProc, x::CuArray)
...
     with_context(from_proc) do
-        CUDA.synchronize()
-        _x = adapt(Array, x)
-        CUDA.synchronize()
-        return _x
+        ev = CUDA.CuEvent()
+        CUDA.record(ev, stream())
+        CUDA.synchronize(ev) # CPU waits ONLY for this stream to finish
+        
+        return adapt(Array, x)
```
jpsamaroo, line 188: I don't think this is necessary - `CUDA.synchronize()` is a stream-local sync (it's the same as `CUDA.synchronize(stream())`.

ext/CUDAExt.jl
```diff
@@ -153,7 +156,6 @@ end
 function Dagger.move(from_proc::CPUProc, to_proc::CuArrayDeviceProc, x)
     with_context(to_proc) do
         arr = adapt(CuArray, x)
-        CUDA.synchronize()
```
jpsamaroo, line 156: These are necessary to ensure that `x` isn't modified before it's read into `arr`, I believe.

ext/CUDAExt.jl
```diff
@@ -200,31 +201,39 @@ function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CPUProc, x::Chunk)
...
 function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CPUProc, x::CuArray{T,N}) where {T,N}
     with_context(from_proc) do
-        CUDA.synchronize()
+        ev = CUDA.CuEvent()
+        CUDA.record(ev, stream())
+        CUDA.synchronize(ev)
+        
         _x = Array{T,N}(undef, size(x))
         copyto!(_x, x)
-        CUDA.synchronize()
```
jpsamaroo, line 206: Same with this function, this shouldn't be necessary.

ext/CUDAExt.jl
```diff
@@ -200,31 +201,39 @@ function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CPUProc, x::Chunk)
...
 function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CuArrayDeviceProc, x::Dagger.Chunk{T}) where T<:CuArray
     if from_proc == to_proc
-        # Same process and GPU, no change
-        arr = unwrap(x)
-        with_context(CUDA.synchronize, from_proc)
-        return arr
+        # Same process and GPU, no change.
+        # Stream ordering guarantees safety; no sync needed.
+        return unwrap(x)
+
```
jpsamaroo, line 215:
```suggestion
```
Please remove these kinds of extra whitespace and newlines, even if Claude/Gemini adds them 😄

ext/CUDAExt.jl
```diff
@@ -200,31 +201,39 @@ function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CPUProc, x::Chunk)
...
+        end
+        
         return with_context(to_proc) do
+            CUDA.wait(ev, stream()) 
             to_arr = similar(from_arr)
             copyto!(to_arr, from_arr)
-            CUDA.synchronize()
             return to_arr
         end
+
```
jpsamaroo:
```suggestion
```

ext/CUDAExt.jl
```diff
@@ -252,14 +261,17 @@ function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CuArrayDeviceProc, x
             end
         else
             return arr
-        end
+        end
```
jpsamaroo:
```suggestion
        end
```

ext/CUDAExt.jl
```diff
@@ -252,14 +261,17 @@ function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CuArrayDeviceProc, x
...
         # Different node, use DtoH, serialization, HtoD
         host_copy = remotecall_fetch(from_proc.owner, from_proc, x) do from_proc, x
             return with_context(from_proc) do
+                ev = CUDA.CuEvent()
+                CUDA.record(ev, stream())
+                CUDA.synchronize(ev) 
                 Array(unwrap(x))
             end
-        end
+        end
```
jpsamaroo:
```suggestion
        end
```

ext/CUDAExt.jl
```diff
@@ -252,14 +261,17 @@ function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CuArrayDeviceProc, x
...
             return arr
-        end
+        end        
     else
         # Different node, use DtoH, serialization, HtoD
         host_copy = remotecall_fetch(from_proc.owner, from_proc, x) do from_proc, x
             return with_context(from_proc) do
+                ev = CUDA.CuEvent()
+                CUDA.record(ev, stream())
+                CUDA.synchronize(ev)
```
jpsamaroo: This could actually just be done as `CUDA.synchronize()`, we don't need an event here.

ext/CUDAExt.jl
```diff
@@ -268,20 +280,30 @@ end
 
 function Dagger.move(from_proc::CuArrayDeviceProc, to_proc::CuArrayDeviceProc, x::CuArray)
     if from_proc == to_proc
-        with_context(CUDA.synchronize, from_proc)
         return x
+
```
jpsamaroo:
```suggestion
```

ext/CUDAExt.jl
```diff
@@ -268,20 +280,30 @@ end
...
+        end
+        
         return with_context(to_proc) do
+            CUDA.wait(ev, stream())
             to_arr = similar(x)
             copyto!(to_arr, x)
-            CUDA.synchronize()
             return to_arr
         end
+
```
jpsamaroo:
```suggestion
```

ext/CUDAExt.jl
```diff
@@ -268,20 +280,30 @@ end
...
             copyto!(to_arr, x)
-            CUDA.synchronize()
             return to_arr
         end
+        
     else
         host_copy = with_context(from_proc) do
+            ev = CUDA.CuEvent()
+            CUDA.record(ev, stream())
+            CUDA.synchronize(ev)
```
jpsamaroo: This one also is probably just `CUDA.synchronize()`

jpsamaroo (approved):
Great work! I have some tweaks for you to make, but this otherwise is looking good!

ext/CUDAExt.jl
victorcamaraa: Yeah, this is a bad habit of mine, I will make sure i remove them next time

PR #702, opened 2026-04-24
add cholesky MPI test

yanzin00:
- added cholesky test using MPI acceleration
- fixed some bugs related to MPI test, like `no method matching accel_matches_proc `


jpsamaroo:
Implemented in #720

PR #704, opened 2026-05-26
DArray/stencil: Fix ROCm GPU CI

jpsamaroo:
Reduces memory usage, which at least for ROCm was causing the OOM killer to start.

Written by Cursor Composer


jpsamaroo:
Moving these commits into https://github.com/JuliaParallel/Dagger.jl/pull/704

PR #720, opened 2026-07-16
MPI: Final merge sprint of the whole MPI infrastructure

fda-tome:
Added the whole SPMD-uniform structure and it's EXT companions for accelerators, as well as the MPI RPC infrastructure that was needed.


jpsamaroo (approved):
Finally! Great work @fda-tome and @yanzin00 :tada:

